PR #168, opened 2021-05-15
Broker/port service

FrozenKiwi:
Porting broker service to use the new shared processor


apps/broker-service/.vscode/launch.json
```diff
@@ -17,10 +17,7 @@
         "ts-node/register",
         "${workspaceFolder}/src/main.ts",
       ],
-      "envFile": "${workspaceFolder}/../../environments/debug.env",
-      "env": {
-        "SETTINGS": "live"
-      },
+      "envFile": "${workspaceFolder}/../../environments/devlive.env",
```
FrozenKiwi, line 20: This project cannot function without a database, and it turns out our mock doesn't run without jest :-(

Maybe what I should do is load jest on startup?

FrozenKiwi: running this without devlive would be way easier...

apps/broker-service/src/controllers/ETransfer.controller.ts
```diff
@@ -36,18 +37,8 @@ export class ETransferController extends Controller {
...
         }
-      }
-      catch (err) {
-        console.error(err.message);
-        SendMail("TransferCode Error", JSON.stringify(err) + "\n---\n" + JSON.stringify(request));
-        return {
-          code: "TheCoin",
-          error: "server Error",
-        }
-      }
```
FrozenKiwi, line 51: although this isn't the point of this PR, our broker already has generalized error handling that does a better job than our custom setups here

apps/broker-service/src/controllers/Newsletter.controller.ts
```diff
@@ -1,94 +1,75 @@
...
+
+// Convert from datetime rep to number timestamp
+function convert(sub?: Subscription) {
+  if (!sub) return { error: "Not Found" };
+  const { registerDate, ...rest } = sub;
+  return {
+    registerDate: registerDate.toMillis(),
+    ...rest,
+  }
+}
```
FrozenKiwi, line 15: Eventually, all our code should use exclusively `DateTime`.  Not `Timestamp`, not `Date`, and not `number`.  Any conversions happen at the very leaves, and nobody else should be aware that they are happening.

MarieTheCoin (approved):

apps/broker-service/.vscode/launch.json
FrozenKiwi: https://github.com/thecointech/thecoin/issues/184

apps/broker-service/src/exchange/VerifiedBillPayments.test.ts
```diff
@@ -1,51 +1,39 @@
...
-    expect(r.hash).toEqual(tx.hash);
-    expect(r.fiatDisbursed).toEqual(0);
-    //expect(r.encryptedPayee).toMatch(name);
-
-    // Now check that we have saved the payee info
-    //const savedPayee = (await GetNamedPayee(wallet.address, name))!;
-    // expect(savedPayee.payee).toEqual(billPayment.encryptedPayee.encryptedPacket);
-    // expect(savedPayee.version).toEqual(billPayment.encryptedPayee.version);
-  })
+it("can process a bill payment", async () => {
```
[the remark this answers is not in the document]
FrozenKiwi: https://github.com/thecointech/thecoin/issues/183

PR #169, opened 2021-05-15
Firestore/data conversion

FrozenKiwi:


libs/broker-db/src/converter.test.ts
```diff
@@ -0,0 +1,59 @@
...
+
+  beforeAll(() => {
+    init();
+  });
+  afterEach(() => {
+    jest.clearAllMocks();
+  })
+
+  // Test Address.  Uses Timestamp to give uniqueness because tests
+  // do not clean up after themselves.
```
FrozenKiwi, line 20: we allow these tests to connect with the emulator (if its running) because I wanted to be able to see what the results actually looked like. The emulator doesn't clean up after itself

libs/broker-db/src/converter.ts
```diff
@@ -0,0 +1,50 @@
...
+    }
+    else if (isPlainObject(obj[k])) {
+      r[k] = convertObject(obj[k], keys, converter);
+    }
+  })
+  return r;
+}
+type converter = {
+  toFirestore: (v: any) => any;
+  fromFirestore: (v: any) => any;
```
FrozenKiwi, line 20: I don't like any's but I couldn't figure out how to do these as a template, and they are not export so the `any` isn't likely to cause confusion

PR #178, opened 2021-05-17
Port/reconcilliation

FrozenKiwi:
Build fixes, with some patching of the tx-reconcilliation project.

This job is very large though, and we are (for now) simply commenting out or bypassing non-functional code.  The refactoring is too large to fit into this PR, and the work is not very high-priority


apps/admin/app/containers/BrokerCAD/VerifyAccount/index.tsx
```diff
@@ -62,8 +63,8 @@ class VerifyAccount extends React.PureComponent<Props, typeof initialState> {
 		const address = NormalizeAddress(account);
 		const signature = await signer.signMessage(address)
 
-		await SetUserVerified(signature, address, Timestamp.now());
-		await CreateReferrer(signature, address);
+		await setUserVerified(signature, address, DateTime.now());
+		await createReferrer(signature, address);
 		alert('Done');
```
MarieTheCoin, line 68: Alert?

MarieTheCoin (approved):

apps/admin/app/containers/BrokerCAD/VerifyAccount/index.tsx
FrozenKiwi: Yeah... it's Admin (internal only).

Still should be better, but...

PR #179, opened 2021-05-18
Right column out of app

MarieTheCoin:


apps/site-app/src/components/AppContainers/index.tsx
```diff
@@ -3,17 +3,18 @@ import styles from './styles.module.less';
 
 type AppContainerProps = {
   className?: string,
+  id?: string
 }
 
 export const AppContainer : React.FC<AppContainerProps> = (props) =>
-  <div className={ `${styles.appContainer} ${props.className}` }>{props.children}</div>;
+  <div id={ `${props.id}` } className={ `${styles.appContainer} ${props.className}` }>{props.children}</div>;
```
MarieTheCoin, line 10: Ended up not using that but will be useful later so I am leaving it

apps/site-app/src/containers/App/index.tsx
```diff
@@ -18,8 +18,6 @@ import { MainRouter } from 'containers/MainRouter';
 import { useFxRatesStore } from '@thecointech/shared/containers/FxRate/reducer';
 import { useSidebar } from '@thecointech/shared/containers/PageSidebar/reducer';
 import { GreaterThanMobileSegment, MediaContextProvider, mediaStyles } from '@thecointech/shared/components/ResponsiveTool';
-import { ColumnRightTop } from 'containers/ColumnRight/Top';
-import { ColumnRightBottom } from 'containers/ColumnRight/Bottom';
```
MarieTheCoin, line 22: Those are now in the contents.

libs/shared/src/containers/Account/Account.tsx
```diff
@@ -23,6 +23,7 @@ export type RouterPath = {
   exact?: boolean;
   icon?: SemanticICONS;
   header?: { avatar: string, primaryDescription: string, secondaryDescription: string | Element | JSX.Element };
+  key?: string
```
MarieTheCoin, line 26: Will be useful leaving it.

FrozenKiwi (approved):
So, I understand what/why you've done it this way.

I disagree with it on a fundamental level: it very much breaks the top-down component-wise structure of the app.  It's also completely the opposite of how the sidebar functions (and they have very similar concepts).  Solving the same job in 2 different ways is the kind of things that makes future dev more difficult.

For now: bad working > good broken, and it's a philosophical issue: so lets submit this PR, but please add an issue to revisit this later.  One day, when we are less stressed, we can find a more holistic approach to this problem

PR #188, opened 2021-05-21
App/hide and show transactions filters

MarieTheCoin:
I tried to fit the balance in the mobile version for the transaction but was not able to make it work properly so I left it for now. I will think of something.

Capture d’écran, le 2021-05-21 à 21 36 56
Capture d’écran, le 2021-05-21 à 21 35 27


libs/shared/src/containers/TransactionList/TransactionLine/index.tsx
```diff
@@ -0,0 +1,71 @@
...
+
+  classForMoneyCell: string,
+  changeCad: number,
+  timeToDisplay: string,
+
+  balanceCad: number
+
+};
+                
+export const TransactionLine = (props:VisualProps) => {
```
MarieTheCoin, line 26: Spend some time debating if I should have split that and I decided that maybe it was too much.

libs/shared/src/containers/TransactionList/index.tsx
```diff
@@ -53,77 +77,75 @@ export const TransactionList = (props: MyProps) => {
...
+        classForMoneyCell={classForMoneyCell}
+        changeCad={changeCad}
+        timeToDisplay={timeToDisplay}
+      
+        balanceCad={balanceCad}
+      />
+  )});
+  const transactionsListZone = txJsxRows.length > 0 
+      ? <Grid padded>{...txJsxRows}{jsxFooter}</Grid> 
+      : <Segment placeholder><Header as="h4" icon><Icon name='search' /><FormattedMessage {...notransactions} /></Header></Segment> ;
```
MarieTheCoin, line 137: Added a "no transactions to display" zone it's always better to have that: users won't wonder if the thing is broken.

libs/shared/src/containers/TransactionList/index.tsx
```diff
@@ -28,9 +51,10 @@ function buildPagination(transactions: Transaction[], maxRowCount: number, curre
     const startRow = currentPage * maxRowCount;
     transactions = transactions.slice(startRow, startRow + maxRowCount);
 
-    return [transactions, (<Table.Footer>
-      <Table.Row>
-        <Table.HeaderCell colSpan='4'>
+    return [transactions, (
+    <Grid>
```
MarieTheCoin, line 55: Switched everything to grids: it's way easier to work with.

libs/site-semantic-theme/src/semantic/thecoin/globals/site.variables
```diff
@@ -137,6 +137,7 @@
 @theCoinPaletteGray1     : #6F6571;
 @theCoinPaletteGray2     : #F2F2F2;
 @theCoinPaletteGray3     : #767676;
+@theCoinPaletteGray4     : #6A6A6A;
```
MarieTheCoin, line 140: A grey that must have fell through a crack.

libs/shared/src/containers/TransactionList/TransactionLine/index.tsx
```diff
@@ -0,0 +1,71 @@
...
+import styles from './styles.module.less';
+import { Grid } from 'semantic-ui-react';
+
+type VisualProps={
+  key: number,
+  locale: string,
+
+  yearToDisplay: number,
+  monthTodisplay: string,
+  dayToDisplay: number,
```
FrozenKiwi, line 11: Why did you in date this way, rather than just passing in a `DateTime`?

libs/shared/src/containers/TransactionList/TransactionLine/index.tsx
FrozenKiwi: Seems fine to me?

libs/shared/src/containers/TransactionList/index.tsx
FrozenKiwi: I think between Grid & Flex display, there isn't really much need for Tables anymore.  Old habits die hard though!

libs/shared/src/containers/TransactionList/index.tsx
```diff
@@ -53,77 +77,75 @@ export const TransactionList = (props: MyProps) => {
...
+    const imgForLine = changeCad < 0 ? iconCard : iconThecoin;
+    const classForMoneyCell = changeCad < 0 ? styles.moneyNegative : styles.moneyPositive;
+    const contentForComment = changeCad < 0 ? <FormattedMessage {...sent} /> : <FormattedMessage {...received} />;
+    const descForComment = changeCad < 0 ? <FormattedMessage {...to} /> : <FormattedMessage {...from} />;
+
+    const monthTodisplay = tx.date.setLocale(locale).monthShort;
+    const yearToDisplay = tx.date.setLocale(locale).year;
+    const dayToDisplay = tx.date.setLocale(locale).day;
+
+    const timeToDisplay = tx.date.setLocale(locale).toLocaleString(DateTime.TIME_SIMPLE);
```
FrozenKiwi, line 112: A lot of these are presentation components (ie, deciding how something is displayed, not what it means).  I would have pushed them all into the `TransactionLine`, basically so the TransactionLine can decide how it shows.  Partly separation-of-concerns, but also mostly because writing a story for the TxLine would be easier if it was in complete control how it displayed

FrozenKiwi: I mean pretty much everything from `imgForLine` down to `timeToDisplay`.  It's all the same data, just deciding what pixels to use

libs/shared/src/containers/TransactionList/index.tsx
FrozenKiwi: Yes!  Very important, cause that's the users first impression!

libs/site-semantic-theme/src/semantic/thecoin/globals/site.variables
FrozenKiwi: hahaha - we've ended up with quite a large pallet :-).

Gray1 is very similar to Gray4.  What would you think if we got rid of one of them?

libs/shared/src/containers/TransactionList/index.tsx
MarieTheCoin: Tables are bad and evil. Tables are bad and evil.

libs/site-semantic-theme/src/semantic/thecoin/globals/site.variables
MarieTheCoin: We will have to test that that could look very bad some places

libs/shared/src/containers/TransactionList/TransactionLine/index.tsx
MarieTheCoin: Wanted to have a very stupid component just for the display

libs/shared/src/containers/TransactionList/TransactionLine/index.tsx
```diff
@@ -0,0 +1,71 @@
+import React from 'react';
+import styles from './styles.module.less';
+import { Grid } from 'semantic-ui-react';
+
+type VisualProps={
+  key: number,
+  locale: string,
```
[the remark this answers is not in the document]
MarieTheCoin: Thought it would be easier for storybook

libs/shared/src/containers/TransactionList/index.tsx
MarieTheCoin: I wanted to keep the component really short and stupid thought it would actually be easier for storybook

PR #212, opened 2021-05-27
Fix web3 create and link account

MarieTheCoin:


FrozenKiwi (approved):
Net change of -1300 LoC... sneaky haha :-D

Good stuff, thanks Marie!

PR #213, opened 2021-05-28
Nft/site

FrozenKiwi:
Very basic site.  At least some of the code is duplicated from `site-app`.

It's not as bad as it looks.  The first 4000 LoC you can skip - its duplicated.  I'll de-duplicate it ASAP, but after publishing main site.

The final 2000 lines have a few interesting things in them - Marie, I think you'll like the new grid layout technique (it makes things like margins way easier :-) ).  It is a _lot_ simpler than the main site, but I think the concept could be extended quite easily.  Something to keep in mind for the next misbehaving element

image


FrozenKiwi:
image
Claim page

.storybook/main.js
```diff
@@ -11,6 +11,20 @@ module.exports = {
...
+    return ({
+    ...options,
+    plugins: [
+      ...options.plugins,
+      [
+        "formatjs",
+        {
+          "idInterpolationPattern": "[sha512:contenthash:base64:6]",
+          "ast": true
+        }
```
FrozenKiwi, line 24: It turns out the `formatMessages` call is basically a placeholder, and a build-tool goes through, add's the ID & removes the unnecessary data from the built code.

I've added it to Storybook and all 3 sites

apps/nft-service/src/nft/contract.ts
```diff
@@ -5,3 +5,13 @@ export async function getContract() {
   const signer = await getSigner("BrokerTransferAssistant");
   return connectNFT(signer);
 }
+
+export async function getMinterAddress() {
```
FrozenKiwi, line 9: This function is to allow both production to have a static, pre-defined address, but for dev & devlive to be able to use whatever addresses are current in the emulators

apps/nft-service/src/nft/minter.json
```diff
@@ -0,0 +1,3 @@
+{
+  "address": "0x123123"
```
FrozenKiwi, line 2: TODO, obvs :-)

apps/site-nft/src/api/mock/accounts.ts
```diff
@@ -0,0 +1,67 @@
+import { AccountMapState, buildNewAccount, IAccountStoreAPI } from '@thecointech/shared/containers/AccountMap';
+import testWallet from './testAccount1.json';
+import Thisismy from './Thisismy.wallet.json';
+import { Wallet } from 'ethers';
+import { AccountName, getSigner } from '@thecointech/accounts';
+import { TheSigner } from '@thecointech/shared/SignerIdent';
+import { connectNFT } from '@thecointech/nft-contract';
+
+export const wallets = [
```
FrozenKiwi, line 9: Identical duplicate of what's in `site-app`

apps/site-nft/src/components/AccountSwitcher/AccountSwitcher.stories.tsx
```diff
@@ -0,0 +1,37 @@
...
+
+export default {
+  title: 'NFT/AccountSwitcher',
+  component: AccountSwitcherComponent,
+} as Meta;
+
+const Template: Story = () => <AccountSwitcherComponent />;
+
+export const NoAccounts = Template.bind({});
+NoAccounts.decorators = [
```
FrozenKiwi, line 15: still identical to app

apps/site-nft/src/components/App/index.tsx
```diff
@@ -0,0 +1,73 @@
...
+export const App = () => {
+
+  if (process.env.NODE_ENV === 'development') {
+    //useDevPrep(account)
+    if (process.env.CONFIG_NAME === 'devlive') {
+      const accountsApi = useAccountStoreApi();
+      React.useEffect(() => { addDevLiveAccounts(accountsApi) })
+    }
+  }
+
```
FrozenKiwi, line 35: This is where things start deviating from the original.  This app is structured a little bit differently from the original.  Here we trial using the css `grid` layout instead of the div's in the original.  The content is structured in 3 parts: top/middle/botton.

apps/site-nft/src/components/App/styles.module.less
```diff
@@ -0,0 +1,24 @@
...
+#app {
+  display: grid;
+  grid-template-columns: [full-start] auto [page-start] 173px [content-start] minmax(auto, 837px) [content-end] 173px [page-end] auto [full-end];
+  grid-template-rows: [header-start] 60px [header-end content-start] 1fr [content-end footer-start] 32px [footer-end];
+
+  grid-template-areas:
+    "lgap header header header rgap"
+    "lgap left   main   right  rgap"
+    "lgap footer footer footer rgap";
+}
```
FrozenKiwi, line 12: This is working really well, it defines the entire layout of the app in one spot. It allows for resizable horizontal & vertical, overflow for the header/footer, but keeps everything consistent with each other.

Note: You can also specify grid guttering, which I think would make the margin issues a lot easier to deal with, but I couldn't figure out how to get the guttering to work nicely with dropshadows

apps/site-nft/src/components/App/styles.module.less
```diff
@@ -0,0 +1,24 @@
...
+
+.contentContainer{
+  grid-area: main;
```
FrozenKiwi, line 15: Here we tell the content to inhabit the 'main' grid area

apps/site-nft/src/components/App/styles.module.less
```diff
@@ -0,0 +1,24 @@
...
+}
+
+#mainContent{
+  // Explicitly define a min-height to prevent our content
+  // resizer from collapsing down if our content is too small
+  min-height: 300px;
+  margin: 28px;
```
FrozenKiwi, line 22: Very frustrating that we have to have margins on each element.  Would be very nice to set them once & forget

apps/site-nft/src/components/AppContainers/index.tsx
```diff
@@ -0,0 +1,18 @@
+import React from "react";
+import styles from './styles.module.less';
+
+type AppContainerProps = {
+  className?: string,
+  shadow?: boolean,
+  noPadding?: boolean,
+}
+
+export const AppContainer : React.FC<AppContainerProps> = (props) =>
```
FrozenKiwi, line 10: I refactored AppContainer to just be a single type with arguments, instead of 3 different classes. It makes the composition a bit easier (no AppContainerWithShadowNoPadding)

apps/site-nft/src/components/AuthRoute/index.tsx
```diff
@@ -0,0 +1,59 @@
...
+import { FormattedMessage } from 'react-intl';
+import { useAccount } from '@thecointech/shared/containers/Account';
+
+const waitingForWeb3 = {
+  defaultMessage: "Connecting to your Web3 provider",
+  description:"Message to display while waiting for user to complete Web3 connection"
+};
+
+// An authorized route is one that requires the use of an unlocked account
+export const AuthRoute = (routeProps: RouteProps)  => {
```
FrozenKiwi, line 15: This is the other thing that is significantly different from app.  I originally built the `Account.tsx` class to handle all the account management.  We passed it in a map of all the different children.

Unfortunately, it didn't place nicely with this app, because it had a lot of opinions (ie - it needed a sidebar).  It was doing a lot of jobs at once, and every time I went back to it it felt very confusing.

This component does nearly the same thing, but as an `AuthRoute` attached to a switcher.  It just makes sure that the account is valid before showing the component.  It's a lot nicer to reason about, because it doesn't do tonnes of weird stuff like setup menus.

apps/site-nft/src/components/AuthRoute/index.tsx
```diff
@@ -0,0 +1,59 @@
...
+      return <FormattedMessage {...waitingForWeb3} />
+    }
+  }
+
+  return (
+    <Route {...routeProps} />
+  )
+}
+
+// const connectSigner = async (accountState: AccountState, accountActions: IActions) => {
```
FrozenKiwi, line 49: TODO: validate with Metamask

apps/site-nft/src/components/Footer/styles.module.less
```diff
@@ -0,0 +1,20 @@
+@import "~@thecointech/site-semantic-theme/variables.less";
+
+#footerContainer{
+  grid-area: footer;
```
FrozenKiwi, line 4: Look! The footer requests where it should be displayed, and the app pops it in there.

No need to worry about positioning, it's parents or anything like that!  It's really quite clean!

apps/site-nft/src/components/MainNavigation/styles.module.less
```diff
@@ -0,0 +1,33 @@
+@import "~@thecointech/site-semantic-theme/variables.less";
+
+.background {
+  background-color: white;
+  grid-row-start: header;
+  grid-row-end: header;
+  grid-column-start: lgap;
+  grid-column-end: rgap;
+  filter: drop-shadow(0px 4px 4px rgba(0, 0, 0, 0.25));
```
FrozenKiwi, line 9: Here, we had to get a little more specific with the grid, saying precisely where we start and where we end.

However, this is fine, because with this we still ensure that we are in the right place, and that our content (the menu itself) is aligned with the content.

apps/site-nft/src/components/MainRouter/index.tsx
```diff
@@ -0,0 +1,36 @@
...
+    profile: ProfileBuilder,
+  },
+  Public: {
+    gauth: GAuth,
+    addAccount: AddAccount,
+    congrats: Congratulations,
+  },
+  default: HomePage,
+}
+
```
FrozenKiwi, line 27: So this is an incomplete concept.  I wanted to have a system where we could define all routes in a way that the rest of the site could refer to it (ie - to use the compiler to catch any 404's).  However, it didn't quite work because the Auth path needs a bit more thought.

I originally thought it would be great to have a very black-box system, where the each page would define it's own dependents, but this hasn't worked out: in the end I think it's better to have a single-source-of-truth as to what our site-map looked like.

apps/site-nft/src/components/ProfileBuilder/Editor/styles.module.less
```diff
@@ -0,0 +1,14 @@
+#wrapper {
+
+  width: 100%;
+  :global(.tui-image-editor-header-logo) {
```
FrozenKiwi, line 4: Easiest way to tweak the image editor - just override it's styles.

apps/site-nft/src/i18n.ts
```diff
@@ -0,0 +1,29 @@
...
+  require('@formatjs/intl-pluralrules/dist/locale-data/en'); // Add locale data for en
+  require('@formatjs/intl-pluralrules/dist/locale-data/fr'); // Add locale data for fr
+}
+
+interface Messages {
+  [index:string]: string;
+}
+// eslint-disable-next-line @typescript-eslint/no-var-requires
+const enTranslationMessages: Messages = require('./translations/en.json');
+const frTranslationMessages: Messages = require('./translations/fr.json');
```
FrozenKiwi, line 18: This i18n file is significantly simpler than the original.  I'm not entirely sure, but I think most of what the original code did is no longer necessary

libs/nft-contract/migrations/3_dev_initialize.ts
```diff
@@ -0,0 +1,32 @@
...
+      const minterAddress = await minter.getAddress();
+      const contract = artifacts.require("TheCoinNFT");
+      const nft = await contract.deployed();
+      // Mint 10 NFT's from 10-20 and print out a claim code for each
+      const ids = Array.from({length: 10}, (_, i) => i + 10)
+      await nft.bulkMinting(ids, 2022, {from: minterAddress});
+
+      // Transfer #10 to Client1
+      const client1 = await getSigner("client1");
+      nft.transferFrom(minterAddress, await client1.getAddress(), 10, {from: minterAddress});
```
FrozenKiwi, line 20: Client1 seeded a token by default in `devlive`

libs/nft-contract/src/connect.ts
```diff
@@ -1,21 +1,30 @@
 import { Signer, Wallet } from "ethers";
 import type { TheCoinNFT } from "./types/TheCoinNFT";
 import { getContract } from "./contract";
+import { Network } from 'ethers/utils';
 
-export async function connectNFT(signer: Signer) : Promise<TheCoinNFT> {
-	const contract = await getContract();
-	if ((signer as Wallet).connect) {
+export function connectNFT(signer: Signer) : TheCoinNFT {
```
FrozenKiwi, line 6: By making this function sync, we can now use it during initialization

We'll probably want to do the same to TC contract.  Initialization is still painful

libs/nft-contract/src/contract.ts
```diff
@@ -1,35 +1,40 @@
...
-export async function getContract() : Promise<TheCoinNFT> {
-  globalThis.__contractNFT = globalThis.__contractNFT ?? await buildContract();
+export function getContract() : TheCoinNFT {
+  globalThis.__contractNFT = globalThis.__contractNFT ?? buildContract();
   return globalThis.__contractNFT!;
 }
+
+//
+// A simple workaround to force the TS compiler to copy the file
+export async function forceCompilerToCopyFile() {
```
FrozenKiwi, line 37: note to self: I'm pretty sure this workaround didn't actually work

libs/nft-contract/src/tokenCodes.test.ts
```diff
@@ -0,0 +1,15 @@
+import { Wallet } from 'ethers';
+import { getTokenClaimCode, getTokenClaimAuthority } from './tokenCodes';
+
+it ('can generate a token code', async () => {
+  const auth = Wallet.createRandom();
+  const code = await getTokenClaimCode(10, auth);
+  expect(code.length).toBe(89);
+})
```
FrozenKiwi, line 8: yay tests

libs/site-base/src/components/MainPageTransition/index.tsx
```diff
@@ -2,24 +2,11 @@ import * as React from 'react';
...
-      </ContentHeightAnimate>
-    );
-  }
-}
-
-export default MainPageTransition;
+export const MainPageTransition: React.FC = (props) => (
+  <ContentHeightAnimate>
+    <ContentFader>
+      <ContentHeightMeasure>{props.children}</ContentHeightMeasure>
```
FrozenKiwi, line 9: See - isn't less code nice?

apps/site-nft/src/.htaccess
```diff
@@ -0,0 +1,52 @@
+<ifModule mod_rewrite.c>
+
+
+  #######################################################################
+  # GENERAL                                                             #
+  #######################################################################
+
+  # Make apache follow sym links to files
+  Options +FollowSymLinks
```
MarieTheCoin, line 9: Do we need FollowSymLinks? It could be a security risk.

MarieTheCoin: I don't know if it's possible here but we could also add ServerSignature at Off and ServerTokens Prod. If those are still called that way

apps/site-nft/src/components/AddAccount/Connect/index.tsx
```diff
@@ -0,0 +1,53 @@
+import React, { useState } from 'react';
+import { Form } from 'semantic-ui-react';
+import { ConnectWeb3 } from '@thecointech/shared/containers/Account/Web3';
```
MarieTheCoin, line 3: You might want to catch the latest version of all those files. The changes are mainly styles so not a big deal

apps/site-nft/src/components/HomePage/index.tsx
```diff
@@ -0,0 +1,8 @@
+import { AppContainer } from 'components/AppContainers';
+import * as React from 'react';
+
+export const HomePage = () => (
+  <AppContainer shadow>
+    HOME PAGE
```
MarieTheCoin, line 6: I am feeling anxious seeing text not being translatable even if I know its temporary 😂

apps/site-nft/src/index.html
```diff
@@ -0,0 +1,31 @@
...
+  <meta charset="utf-8">
+
+  <!-- Make the page mobile compatible -->
+  <meta name="viewport" content="width=device-width, initial-scale=1">
+
+  <!-- Allow installing the app to the homescreen -->
+  <meta name="mobile-web-app-capable" content="yes">
+
+  <link rel="icon" href="/favicon.ico" />
+  <title>TheCoin NFT - Growing a better future</title>
```
MarieTheCoin, line 17: Non translatable text 😭

apps/site-nft/src/translations/en.json
```diff
@@ -0,0 +1,43 @@
+{
+  "'/BZiQv'": "Upload Failed",
+  "'/lGbT2'": "Load an account stored online",
```
MarieTheCoin, line 3: This is going to be complicated for translation with no context.

MarieTheCoin (approved):
The only problems I can see are the htaccess and the files with spaces in their names (not sure if/how those are used).
Also I didn't know you were an artist. That green circle is great: is it a tree?

apps/nft-service/.vscode/launch.json
```diff
@@ -17,7 +17,7 @@
         "ts-node/register",
         "${workspaceFolder}/src/main.ts"
       ],
-      "envFile": "${workspaceFolder}/../../environments/debug.env",
+      "envFile": "${workspaceFolder}/../../environments/devlive.env",
```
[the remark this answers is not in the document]
FrozenKiwi: Well... debatable.  Probably not - it probably should be `development.env`.  At the moment though, I'm working in devlive to get the systems running nicely, and I need my working breakpoints!

apps/site-app/package.json
```diff
@@ -52,7 +52,6 @@
   },
   "dependencies": {
     "@artsy/fresnel": "^1.2.2",
-    "@formatjs/intl-pluralrules": "^1.4.1",
```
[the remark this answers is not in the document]
FrozenKiwi: Site base depends on it, so we don't need the additional dependency here.  Having it  defined in multiple places just risks problems if different projects depend on different versions.

apps/site-app/src/api/mock/accounts.ts
```diff
@@ -50,16 +50,16 @@ export function getDevWallets() : AccountMapState {
...
-async function addRemoteAccount(name: AccountName, accountsApi: IAccountStoreAPI) {
+async function addRemoteAccount(name: AccountName, accountsApi: IAccountStoreAPI, active: boolean) {
   const signer = await getSigner(name)
   const address = await signer.getAddress();
   const theSigner = signer as TheSigner;
   theSigner.address = address;
   theSigner._isSigner = true;
 
-  accountsApi.addAccount(name, theSigner, false, true);
+  accountsApi.addAccount(name, theSigner, false, active);
```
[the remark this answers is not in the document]
FrozenKiwi: Not sure what you mean by that - this just decides which one is active (same as switching accounts in the account switcher)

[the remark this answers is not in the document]
FrozenKiwi: This is actually mostly for NFT site, because only Client1 has a token, and it was annoying having to switch active account every time I wanted to test signing an NFT

apps/site-nft/src/.htaccess
FrozenKiwi: Ha!  That's where that error message comes from!

This .htaccess file is not actually used (it's not used on Firebase Hosting, which is the hosting we will almost certainly go with). 

I'll delete the whole file

apps/site-nft/src/components/HomePage/index.tsx
FrozenKiwi: Sorry!  The other pages will hopefully make you feel a bit better!

apps/site-nft/src/index.html
FrozenKiwi: OOOoohhhhh!!!!

Good catch!  This is on all our sites... how do we do the translation for an element outside react?

Will log a bug!  May require some investigation

FrozenKiwi: https://github.com/thecointech/thecoin/issues/214

apps/site-nft/src/translations/en.json
FrozenKiwi: It needs it's descriptions.  It also needs to be in site-base, so we don't duplicate the translations.

https://github.com/thecointech/thecoin/issues/215

apps/site-nft/src/index.html
FrozenKiwi: https://github.com/thecointech/thecoin/issues/216

apps/site-nft/src/translations/en.json
FrozenKiwi: https://github.com/thecointech/thecoin/issues/216

PR #217, opened 2021-05-31
Nft/Sign & Upload

FrozenKiwi:
Working complete upload pathway

Unfortunately no mocked IPFS, as there are no current mocks for this project.  This means that dev:live still connects to live server, so cannot be used without API keys.

This PR is mostly bug-fixing to existing code as we finished off testing the IPFS integration


apis/broker/ts/.openapi-generator/VERSION
```diff
@@ -1 +1 @@
-5.0.1
\ No newline at end of file
+5.1.1
```
FrozenKiwi, line 1: I did an investigation on using a generic IPFS API (instead of the specific Pinata API) with the hope it would be able to connect to a mocked IPFS node.

Unfortunaltey, I misunderstood the whole purpose of the mocked node and the IPFS API - it wasn't for uploading, it was to mock pinning (expecting that a local IPFS node would have already uploaded the actual data)

But since it's here, we might as well leave hte upgrade in because it has fixes for the version of TS we are using

apis/broker/ts/common.ts
```diff
@@ -91,7 +91,14 @@ export const setSearchParams = function (url: URL, ...objects: any[]) {
     const searchParams = new URLSearchParams(url.search);
     for (const object of objects) {
         for (const key in object) {
-            searchParams.set(key, object[key]);
+            if (Array.isArray(object[key])) {
+                searchParams.delete(key);
+                for (const item of object[key]) {
+                    searchParams.append(key, item);
+                }
+            } else {
```
FrozenKiwi, line 99: auto-generated code (should not actually be in Git)

apps/nft-service/src/nft/contract.ts
```diff
@@ -8,8 +8,9 @@ export async function getContract() {
 
 export async function getMinterAddress() {
   if (process.env.NODE_ENV === 'production') {
-    const {address} = await import('./minter.json');
-    return address;
+    if (!process.env.WALLET_NFTMinter_ADDRESS)
+      throw new Error('Minter Address no specified');
+    return process.env.WALLET_NFTMinter_ADDRESS
```
FrozenKiwi, line 13: Continuing to move config vars into env.

apps/nft-service/src/nft/ipfs.ts
```diff
@@ -3,19 +3,21 @@
...
 // Upload buffer to IPFS, return upload hash
-export async function upload(file: Buffer) {
+export async function upload(file: Buffer, filename: string) {
   log.trace(`Uploading ${file.byteLength} bytes`);
-
   const stream = Readable.from(file);
+
+  // Fixes upload issue: https://github.com/PinataCloud/Pinata-SDK/issues/28
+  // ¡¡ THE HACK !!
+  (stream as any).path = filename;
```
FrozenKiwi, line 20: The filename is supposed to be irrelevant, as IPFS addresses are content-dependent.  This API is only a few months old though.  Yay for the bleeding edge.

libs/nft-contract/src/deployed/prod.json
```diff
@@ -0,0 +1 @@
+{"contract":"0x2d00"}
```
FrozenKiwi, line 1: These variables are generated at build time, so, unfortunately, cannot be moved to env.vars

apps/nft-service/src/nft/ipfs.ts
```diff
@@ -31,7 +33,7 @@ export async function uploadAvatar(avatar: Buffer, signature: string) {
   // First, validate we are uploading something that is probably right.
   if (!await validateImage(avatar, signature))
     return null;
-  return upload(avatar);
+  return upload(avatar, "avatar.jpg");
```
MarieTheCoin, line 36: Is it ok if it overrides the last file imported?

MarieTheCoin (approved):

apps/nft-service/src/nft/ipfs.ts
FrozenKiwi: I tested it, and there are multiple results on the IPFS node.

I -think- it would be better to have a unique filename, but then nobody (other than us) should ever see these names.  And actually, I'm not sure being able to identify the origin of any given image is a good thing (like, maybe that's too much info?  Could then connect a users Eth account with their profile pic, then their real-world identity, then connect their eth spending etc.  In the end, it may actually be the better choice to strip away any other real-world info.

image

PR #218, opened 2021-06-01
Validation page for images

FrozenKiwi:
Super-super quick page to validate images.

This is in no way complete.  It does not check any actual data, just that the owner can sign images, and this one was signed.  It's primary reason is to give Valeriya an idea to reference against.


apps/site-nft/src/components/Validate/index.tsx
```diff
@@ -0,0 +1,103 @@
...
+import { log } from '@thecointech/logging';
+import icon from './images/icon_topup_big.svg';
+import { Upload } from '../Upload';
+import { getHash, readExif } from '../ProfileBuilder/SignAndUpload/sign';
+import { utils  } from 'ethers';
+import { getContract } from '@thecointech/nft-contract';
+
+const title = defineMessage({ defaultMessage: "Validate Image", description: "page title" });
+const description = defineMessage({ defaultMessage: "Validate the images cryptographic signature", description: "page instructions" });
+//const imageUri = defineMessage({ defaultMessage: "URI of image to validate", description: "Optional uri" });
```
FrozenKiwi, line 15: I took this back out because it's not worth hooking up.  We will eventually use it though, so only commented

environments/devlive.env
```diff
@@ -16,8 +16,8 @@ IPFS_EMULATOR=8378
 
 # Default IPFS access key.  These values must match the
 # the token secret passed to mock-ipfs-pinning-service
-PINATA_API_KEY=ipfskey
-PINATA_API_SECRET=ipfspwd
+PINATA_API_KEY=36d74aea0159902a10df
+PINATA_API_SECRET=9fd41541e589b957756850b18098e459fedfe29439e5f6a888f65ace5c65ae6d
```
FrozenKiwi, line 20: faaarrrrr....

Put this in (just while testing) here because we (currently) don't have a mock for IPFS.  Thats what I get for cutting corners

Reverted in the next PR, and keys revoked,

MarieTheCoin (approved):

PR #219, opened 2021-06-01
Nft/offset listing

FrozenKiwi:
A sample NFT offset listing page.  This is a very quick example page - mostly to give the designer something to reference.  Obvs none of the data on here is real

NOTE: based on validate branch.  To remove that code from this review, set base to not-dev, then back to dev to reset once validate PR is complete

image


apps/nft-service/src/nft/ipfs.test.ts
```diff
@@ -13,7 +13,7 @@ describe('Pinata IPFS integration', () => {
     const f = readFileSync(__filename);
     // Dynamic import so it picks up the env vars set above
     const { upload } = await import('./ipfs');
-    const r = await upload(f, "testf.txt");
```
FrozenKiwi, line 16: test filename overwrite

apps/site-nft/src/components/Offsets/index.tsx
```diff
@@ -0,0 +1,62 @@
...
+import randomhex from 'randomhex';
+import styles from './styles.module.less'
+
+const messages = defineMessages({
+  title: { defaultMessage: "NFT Offsets", description: "NFT Offsets listings page title" },
+  description: { defaultMessage: "Review the offsets each NFT has completed", description: "NFT Offsets listing description" },
+})
+
+const avatars = [
+  'https://react.semantic-ui.com/images/avatar/small/rachel.png',
```
MarieTheCoin, line 16: in the shared components we have our icons/avatars

MarieTheCoin (approved):

apps/site-nft/src/components/Offsets/index.tsx
FrozenKiwi: that would have been better - I didn't even think to look.

This is just throw-away code - so I will leave it like this if that's OK?

PR #222, opened 2021-06-04
Build/cleanup

FrozenKiwi:
Cleanup our build commands, and remove a lot of auto-generated code.

The build command was overly complex.  This PR takes simplifies the commands and eliminates the manual dependency resolution so we can just rely on Lerna as our build runner.

NOTE: 25K LoC down.  Not a bad day :-).


apis/broker/ts/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@thecointech/broker-cad",
-  "version": "0.2.4",
```
FrozenKiwi, line 3: I don't know why this one went backwards, but as we don't publish this package (it's only used internally) it kinda doesn't matter.

apis/nft/ts/package.json
```diff
@@ -0,0 +1,27 @@
...
+  "description": "OpenAPI client for @thecointech/nft",
+  "author": "OpenAPI-Generator Contributors",
+  "keywords": [
+    "axios",
+    "typescript",
+    "openapi-client",
+    "openapi-generator",
+    "@thecointech/nft"
+  ],
+  "license": "Unlicense",
```
FrozenKiwi, line 13: This is the file I think that broke your build the other day.  Was accidentally ignored

apps/site-app/src/app.tsx
```diff
@@ -26,9 +26,8 @@ import { App } from 'containers/App';
 // Import Language Provider
 import { LanguageProvider, Languages } from '@thecointech/shared/containers/LanguageProvider';
 
-// Load the favicon and the .htaccess file
+// Load the favicon
 import '!file-loader?name=[name].[ext]!./images/favicon.ico';
-import 'file-loader?name=.htaccess!./.htaccess';
```
FrozenKiwi, line 31: there is no .htaccess on Firebase.  No point keeping it in our code base in that case.

package.json
```diff
@@ -20,13 +20,8 @@
     "gen:broker": "openapi-generator-cli generate -i apps/broker-service/src/api/swagger.json -g typescript-axios -o apis/broker/ts -c apis/broker/client-ts-spec.json",
     "gen:nft": "openapi-generator-cli generate -i apps/nft-service/src/api/swagger.json -g typescript-axios -o apis/nft/ts -c apis/nft/client-ts-spec.json",
     "genapis": "lerna run gen:swagger && run-p gen:*",
-    "build:libs": "lerna run build:lib",
-    "build:apps": "lerna run build:app --stream",
-    "build:apis": "run-p build:api:*",
```
FrozenKiwi, line 25: Our root `build` command.  Much simpler :-)

MarieTheCoin (approved):

apis/nft/ts/package.json
FrozenKiwi: Remind me of this on Monday.  I'm not sure what you mean, but it obvs needs to be fixed!

apps/rates-service/tsoa.json
```diff
@@ -20,7 +20,7 @@
   ],
   "compilerOptions": {
     "paths": {
-      "@thecointech/utilities/*":  ["../../libs/utils-ts/build/*"]
+      "@thecointech/utilities/*":  ["../../libs/utils-ts/src/*"]
```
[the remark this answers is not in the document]
FrozenKiwi: So - I mean, it's not pretty, but it allows us to run `generate` before we actually build.  This step just needs the typings, so it doesn't actually want the compiled code, so linking it to src is (I think) fine.

It does kinda seem like the thing that will bite us later though

PR #223, opened 2021-06-07
fixed the build, fixed services failing to start on dev:live

FrozenKiwi:


.gitignore
```diff
@@ -20,6 +20,9 @@
 **/*secret*
 **/.firebase/hosting.*.cache
 
+# ignore the IPFS instance we spin up in dev:live
```
MarieTheCoin, line 23: Comments!!! 👌

package.json
```diff
@@ -123,7 +125,7 @@
     "chalk": "^2.4.2",
     "concurrently": "^5.1.0",
     "cpx": "^1.5.0",
-    "cross-env": "5.2.0",
+    "cross-env": "^7.0.3",
```
MarieTheCoin, line 128: Big version change

MarieTheCoin (approved):

package.json
FrozenKiwi: yeah, I'm actually not sure what happened here.  Some kind of merge conflict meant it got uninstalled, so I just re-installed at latest.

It's job is pretty simple: just setting env vars on startup, so... *shrugs*.

FrozenKiwi:
Ignoring snyk - I've tried rebuilding multiple times and it has not removed that error.

PR #224, opened 2021-06-07
FAQ with Prismic

MarieTheCoin:
Capture d’écran, le 2021-06-07 à 18 46 30
Capture d’écran, le 2021-06-07 à 18 46 15
Capture d’écran, le 2021-06-07 à 18 47 28


apps/site-landing/src/containers/HelpDocs/FaqMenu.tsx
```diff
@@ -0,0 +1,47 @@
...
+import { List, Rail } from "semantic-ui-react";
+import { GreaterThanMobileSegment, MobileSegment } from "@thecointech/shared/components/ResponsiveTool";
+
+type Props = {
+    categories: any
+  }
+  
+function displayEntry(index: number, name: string ){
+    return <List.Item key={index}><Link to={"/faq/theme-"+((name)?.split("-"))[0].replace(/ /g, '')}>{((name.split("-"))[1])}</Link>
+                { /* <List> */ }
```
MarieTheCoin, line 13: I thought it could be better to have the list of questions under the FAQ categories in the menu but I wanted to see that with you first.

apps/site-landing/src/containers/HelpDocs/Welcome.tsx
```diff
@@ -1,32 +1,26 @@
...
-
-  const starred =  props.faqs
-    .filter(faq => faq.data.starred)
-
+  const starred =  props.faqs.filter(faq => faq.data.show_on_faq_home);
   return (
-    <>
-      <Header as="h1">
+    <div>
+      <Header as="h2" className={"x10spaceBefore"}>
```
MarieTheCoin, line 18: Following the style from the design

apps/site-landing/src/containers/HelpDocs/index.tsx
```diff
@@ -1,21 +1,20 @@
...
 import { Welcome } from "./Welcome";
+import { selectLocale } from '@thecointech/shared/containers/LanguageProvider/selector';
+import { DEFAULT_LOCALE } from '@thecointech/shared/containers/LanguageProvider/types';
+
 
-const FAQS_KEY = "faqItems";
 
 const HelpDocsInternal = (props: RouteComponentProps) => {
   const actions = usePrismicActions();
-  const sidebarActions = useSidebar();
```
MarieTheCoin, line 18: No more sidebar here 👌

apps/site-landing/src/containers/HelpDocs/index.tsx
```diff
@@ -76,15 +50,19 @@ export const HelpDocs = (props: RouteComponentProps) => {
...
   const { faqs } = docs;
+  const { locale } = useSelector(selectLocale);
   const categories: Dictionary<FAQDocument[]> = {};
   for (const faq of faqs) {
-    const cat = faq.data.category;
-    if (!categories[cat])
-      categories[cat] = [faq]
-    else
-      categories[cat].push(faq);
+    const lang = ((faq.lang)?.split("-")) ? ((faq.lang)?.split("-"))[0] : DEFAULT_LOCALE;
```
MarieTheCoin, line 58: We only get the articles/categories from the language we are in

apps/site-landing/src/containers/HelpDocs/FaqList.tsx
```diff
@@ -0,0 +1,34 @@
...
+  return (
+    <>
+    <div className={styles.containerFaq}>
+      <FaqMenu categories={menu} />
+      <Header as="h2" className={"x10spaceBefore"}>
+        <Header.Content>
+          {title ? ((title.split("-"))[1]) : ""}
+        </Header.Content>
+      </Header>
+      {faqs.filter(faq => locale === ((faq.lang!).split("-"))[0]).map(faq => (<FaqItem key={faq.id} {...faq} />))}
```
FrozenKiwi, line 29: When might this `lang` assertion here be wrong?  Could we just change the type of it to `string` (so we don't need the `!`)

Also, a general style note, it's a lot easier to have only one root-level scope on a line.  If you split `filter` and `map` onto different lines, it would be *way* easier to see what is happening, and also to know what applies to what

apps/site-landing/src/containers/HelpDocs/FaqList.tsx
```diff
@@ -0,0 +1,34 @@
...
+
+export const FaqList = ({ title, faqs, menu }: Props) => {
+  const { locale } = useSelector(selectLocale);
+  return (
+    <>
+    <div className={styles.containerFaq}>
+      <FaqMenu categories={menu} />
+      <Header as="h2" className={"x10spaceBefore"}>
+        <Header.Content>
+          {title ? ((title.split("-"))[1]) : ""}
```
FrozenKiwi, line 26:
```suggestion
          {title?.split("-")[1] ?? ""}
```
A few less brackets go a long way in comprehension.

(Side note, I've not tested that line, so no promises it even compiles)

apps/site-landing/src/containers/HelpDocs/FaqList.tsx
FrozenKiwi:
```suggestion
      {
        faqs
          .filter(faq => locale === faq.lang?.split("-")[0])
          .map(faq => <FaqItem key={faq.id} {...faq} />)
      }
```

FrozenKiwi: again, no idea if that even compiles.  But as far as I can tell, it does the same thing

apps/site-landing/src/containers/HelpDocs/FaqMenu.tsx
FrozenKiwi: That sounds good, but I'm not sure we would want to do it until we had enough questions, And then we'd want to take it away again when we had too many.

I'm fairly ambivalent.  Wouldn't take it out, wouldn't put it in.

apps/site-landing/src/containers/HelpDocs/FaqMenu.tsx
```diff
@@ -0,0 +1,47 @@
...
+import styles from "./styles.module.less";
+import { List, Rail } from "semantic-ui-react";
+import { GreaterThanMobileSegment, MobileSegment } from "@thecointech/shared/components/ResponsiveTool";
+
+type Props = {
+    categories: any
+  }
+  
+function displayEntry(index: number, name: string ){
+    return <List.Item key={index}><Link to={"/faq/theme-"+((name)?.split("-"))[0].replace(/ /g, '')}>{((name.split("-"))[1])}</Link>
```
FrozenKiwi, line 12: the `split` happening here seems to be a very common theme.

I'm not sure what it is supposed to do.  From my POV, it's a very opaque mechanism for selecting things.  It might pay to create a smaller helper function just to make clear whats happening:

```tsx
<List.Item key={index}>
  <Link to={"/faq/theme-"+getWhatsit(name)')}>
     {getThingy(name)}
  </Link>
</>
```
This has the advantage of moving all the complexity of -how- your selecting your wotsit & thingy away somewhere else, so you don't have to reason about that at the same time as your reasoning about the <list>

apps/site-landing/src/containers/HelpDocs/index.tsx
```diff
@@ -26,48 +25,24 @@ const HelpDocsInternal = (props: RouteComponentProps) => {
...
-        //<Route path={buildUrl("").toString()} exact={true} render={()=> <FAQs faqs={docs.faqs} />} />
-      }
-      <Route path={buildUrl("").toString()} exact={true} render={()=> <Welcome faqs={docs.faqs} />} />
-    </Switch>
+    <>
+      <Switch>
+        {
+          Object.entries(categories)
+            .map((entry, index) => {
+              const url = buildUrl("theme-"+((entry[0])?.split("-"))[0].replace(/ /g,'')).toString()
```
FrozenKiwi, line 39: similar comment to above.  I -think- this is the same select, but it would be nice for it to be de-duplicated

apps/site-landing/src/containers/HelpDocs/messages.ts
```diff
@@ -10,14 +10,6 @@ export const scope = 'app.components.Header';
...
-    defaultMessage: 'What would you like to know?',
-  },
-  subHeader: {
-    id: `${scope}.subHeader`,
-    defaultMessage: 'See what interests you from the menu on the left',
-  },
-  blurb: {
-    id: `${scope}.blurb`,
-    defaultMessage: 'Or, for a quick overview of our company try the below items',
+    defaultMessage: 'FAQ',
```
FrozenKiwi, line 13: feel free to move these messages inside the component file, and delete this `messages.ts` file.  The rest of our code embeds the messages directly where they are used

apps/site-landing/src/containers/HelpDocs/index.tsx
```diff
@@ -76,15 +51,19 @@ export const HelpDocs = (props: RouteComponentProps) => {
...
+  const { locale } = useSelector(selectLocale);
   const categories: Dictionary<FAQDocument[]> = {};
   for (const faq of faqs) {
-    const cat = faq.data.category;
-    if (!categories[cat])
-      categories[cat] = [faq]
-    else
-      categories[cat].push(faq);
+    const lang = ((faq.lang)?.split("-")) ? ((faq.lang)?.split("-"))[0] : DEFAULT_LOCALE;
+    const cat = (locale === DEFAULT_LOCALE) ? faq.data.category : faq.data.fr_category;
```
FrozenKiwi, line 60: couldn't `category` be language-neutral?

FrozenKiwi: it sounds like meta-data, which wouldn't be something I'd normally translate

apps/site-landing/src/containers/HelpDocs/styles.module.less
```diff
@@ -0,0 +1,28 @@
...
+
+.containerFaq{
+    width: 50%;
+    max-width: 600px;
+    margin: auto;
+}
+
+#menuFaq{
+    a {
+        color: @theCoinPaletteGreen5;
```
FrozenKiwi, line 11: we need to change the default styling of links - rather than setting then removing styling like this

apps/site-landing/src/containers/MainNavigation/MainNavigationMobile/index.tsx
```diff
@@ -7,19 +7,21 @@ import { Link, NavLink } from 'react-router-dom';
...
-const wedomore = {  id:"site.MainNavigation.wedomore",
+const wedomore = {  id:"site.MainNavigationMobile.wedomore",
                     defaultMessage:"We do more",
                     description:"Title for the We do more entry in the menu"};
-const yourbenefits = {  id:"site.MainNavigation.yourbenefits",
+const yourbenefits = {  id:"site.MainNavigationMobile.yourbenefits",
                     defaultMessage:"Your benefits",
                     description:"Title for the Your benefits entry in the menu"};
-
+const faq = {  id:"site.MainNavigationMobile.faq",
```
FrozenKiwi, line 22: Side note: this text probably shouldn't be duplicated - unless we have different menu entries for mobile/desktop (?)

FrozenKiwi (commented):
This looks pretty good, although I am kinda hoping there is a way to get rid of the localized `category` meta on the FAQ's

FrozenKiwi (approved):

PR #225, opened 2021-06-09
Blog with Prismic + Clean up

MarieTheCoin:
https://user-images.githubusercontent.com/68251967/122625697-730dcc80-d074-11eb-858a-17f8ea272e8b.mov


.gitignore
```diff
@@ -58,3 +58,4 @@ libs/shared/src/containers/IDX/config.devlive.json
 apps/site-nft/src/translations
 apps/site-nft/lang
 
+environments/devlive.env
```
MarieTheCoin, line 61: Should I remove that?

apps/site-landing/src/containers/HelpDocs/FaqItem.tsx
```diff
@@ -1,10 +1,10 @@
...
 import { FAQDocument } from "components/Prismic/types";
 import { Header } from "semantic-ui-react";
-import { Renderer } from "./Renderer/Renderer";
+import { RichText, RichTextBlock } from "prismic-reactjs";
 
 export const FaqItem = ({ data }: FAQDocument) =>
   <div className={"x6spaceBefore x6spaceAfter"}>
-    <Header as={"h3"}><Renderer r={data.question} /></Header>
-    <Renderer r={data.answer} />
+    <Header as={"h3"}>{data.question ? data.question[0].text : ""}</Header>
```
MarieTheCoin, line 8: We don't want people to be able to change the size of the title that would break the design

.gitignore
FrozenKiwi: yes please.  Our devlive should contain no secrets (and anyone should be able to run it)

apps/site-landing/src/components/Prismic/types.ts
```diff
@@ -4,22 +4,40 @@ export type RenderableType = {
...
+}
+
+export type AlternateLang = { 
+  id: string, 
+  type: string, 
+  lang: string 
+}
+
 export type ArticleResult = {
+  thumbnail: any,
```
FrozenKiwi, line 32: Isn't this `any` an `ImageObj`?

apps/site-landing/src/components/Renderer/ElementRender.tsx
```diff
@@ -0,0 +1,24 @@
...
+  {
+    case "heading1":
+      return <Header as={"h1"}>{text}</Header>;
+    case "heading2":
+      return <Header as={"h2"}>{text}</Header>;
+    case "heading3":
+      return <Header as={"h3"}>{text}</Header>;
+    case "heading4":
+      return <Header as={"h4"}>{text}</Header>;
+    case "list-item":
```
FrozenKiwi, line 17: Just curious - but why not `<li>`?

apps/site-landing/src/containers/Blog/index.tsx
```diff
@@ -0,0 +1,71 @@
...
+            })
+        }
+        <Route path="/blog/:articleId" render={(articleProps) => <Article {...articleProps} />} />
+        <Route path={("").toString()} exact={true} render={()=> <Welcome articles={docs.articles} menu={categories} />} />
+      </Switch>
+    </>
+  );
+}
+
+// https://github.com/react-boilerplate/redux-injectors/issues/16
```
FrozenKiwi, line 47: I don't think this issue is relevant here

FrozenKiwi (approved):

apps/site-landing/src/components/Renderer/ElementRender.tsx
MarieTheCoin: It was that way before and I assume it would be difficult to add the ul before with the li. I hope to be able to remove those I am not sure they are actually used that much

PR #226, opened 2021-06-09
All/fix signing

FrozenKiwi:
The new processor would fail on processing e-Transfers due.

The cause was the same signing error we had with NFT transfers, and so the fix was to simply propagate the use of hte sign function throughout the codebase

Fixes:
 - FormattedMessage id generation (broken by merged upgrade to TS4.2)
 - Downgrade TS to 4.0
 - TheCoin contract typing (copy step must have been lost in merges)
 - Seeding of `TestAccNoT`
 - Auto-execute `broker-service` & `nft-service` on `yarn dev:live` in root
 - NOTE: yarn dev:live within contract folders no longer boots up a blockchain
 - Wrong signature passed to `certifiedTransfer` in `depositCoin.ts`


apps/nft-service/src/nft/validate.ts
```diff
@@ -1,5 +1,5 @@
 import { log } from '@thecointech/logging';
-import { verifyMessage } from 'ethers/utils';
```
FrozenKiwi, line 2: The TS version revert seems to have issues with importing directly from sub-folders.

apps/site-nft/src/api/mock/accounts.ts
```diff
@@ -30,22 +29,25 @@ export const wallets = [
...
+let r: AccountMapState|null = null;
 export function getDevWallets() : AccountMapState {
-  const encrypted = wallets[0];
-  const encryptedAccount = buildNewAccount(encrypted.name, JSON.parse(encrypted.wallet));
-  // We always add one encrypted wallet
-  const r: AccountMapState =  {
-    active: null,
-    map: {
-      [encryptedAccount.address]: encryptedAccount
+  if (!r) {
```
FrozenKiwi, line 34: prevent multiple re-initializations.

Our state initialization -still- needs tweaking

apps/site-nft/src/api/mock/accounts.ts
```diff
@@ -30,22 +29,25 @@ export const wallets = [
...
+      const randomAccount =  buildNewAccount("Random Test", Wallet.createRandom());
+      //randomAccount.contract = connectNFT(randomAccount.signer);
+      r.active = randomAccount.address;
+      r.map[randomAccount.address] = randomAccount
     }
-  }
-  // if dev mode, we add a random wallet,
-  if (process.env.SETTINGS !== 'live') {
-    const randomAccount =  buildNewAccount("Random Test", Wallet.createRandom());
-    randomAccount.contract = connectNFT(randomAccount.signer);
```
FrozenKiwi, line 46: we now correctly type the contract, which means that this call didn't compile.  We need to consider how to handle multiple contracts

package.json
```diff
@@ -171,7 +171,7 @@
     "ts-loader": "^8.0.11",
     "ts-node": "^9.1.1",
     "tsconfig-paths-webpack-plugin": "^3.2.0",
-    "typescript": "^4.2.4",
```
FrozenKiwi, line 174: Required for compatibility with formatJs build tools

MarieTheCoin (approved):

PR #232, opened 2021-06-22
Build/tests

FrozenKiwi:
All tests passing
Adding in CI trigger to run tests.  Currently runs in 'build' step, but we could consider splitting hits into a separate job


apps/rates-service/src/internals/rates/UpdateDb.test.ts
```diff
@@ -1,110 +1,88 @@
...
+// Start 5 mins into the mocked data
+var now = 1593696900000;
+const sixHrs = 6 * 60 * 60 * 1000;
+
+beforeEach(async () => {
+  jest.setTimeout(30000);
+})
+
+it('can update Coin rates', async () => {
+  await init({});
```
FrozenKiwi, line 20: use mocked DB instead of emulator. Much simpler & faster

apps/site-app/jest.config.js
```diff
@@ -1,13 +1,5 @@
-const defaults = require('../../jest.config');
+const defaults = require('@thecointech/site-base/jest.config');
```
FrozenKiwi, line 1: website-style jest configs put into site-base and shared out from there

apps/site-app/src/containers/App/tests/__snapshots__/index.test.tsx.snap
```diff
@@ -1,25 +1,87 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
```
FrozenKiwi, line 2: I suspect we will remove this test fairly quickly, I'm not at all sure it's a good idea to do snapshot testing (yet)

apps/tx-processor/src/deposits/index.test.ts
```diff
@@ -1,15 +1,17 @@
...
+import { RbcApi } from '@thecointech/rbcapi';
 
 it("Can complete deposits", async () => {
 
   jest.setTimeout(900000);
   init({});
 
   const theContract = await GetContract();
-  const deposits = await processUnsettledDeposits(theContract);
+  const bank = new RbcApi();
```
FrozenKiwi, line 13: This is mocked, but I would actually kinda prefer it was clearer.

It makes me feel nervous to see my testing code running with a potential connection to my bank account

environments/development.env
```diff
@@ -8,3 +8,7 @@
 
 NODE_ENV=development
 CONFIG_NAME=development
+
+#
+# We need to keep storage path
+STORAGE_PATH=/temp/TheCoin/tests/dbs/
```
FrozenKiwi, line 14: this storage path is a real pain.  It's been mostly refactored away to be unnecessary, but the remaining reasons are very hard to eliminate

jest.config.js
```diff
@@ -1,15 +1,29 @@
...
-    "^.+\\.tsx?$": "ts-jest"
+    "^.+\\.tsx?$": "ts-jest",
+    "^.+\\.jsx?$": "babel-jest",
+  },
+  globals: {
+    'ts-jest': {
+      tsConfig: {
+        ...compilerOptions,
+        noUnusedLocals: false,
+        noUnusedParameters: false,
```
FrozenKiwi, line 16: note: jest tests are now a lot more permissive about what code they will accept.

No more annoying compiler errors on test code!

libs/logging/src/browser.ts
```diff
@@ -2,12 +2,12 @@ import { createLogger, TRACE, INFO, DEBUG, stdSerializers } from 'browser-bunyan
 import { ConsoleFormattedStream } from '@browser-bunyan/console-formatted-stream';
 import { BunyanLogger } from './logger';
 
-export function init_browser(name: string): BunyanLogger {
+export function init_browser(name: string, level?: number): BunyanLogger {
```
FrozenKiwi, line 5: allows suppressing logging in tests.

libs/store/src/config.test.ts
```diff
@@ -37,32 +30,3 @@ it("Can store KV pairs in Config", async () => {
 
   await ConfigStore.release();
 });
-
-it("Can create DB on disk", async () => {
```
FrozenKiwi, line 41: if we're running on GitHub CI, no writing to disk

libs/tx-gmail/src/auth.ts
```diff
@@ -22,13 +21,10 @@ export type OpenUrl = (url: string) => void;
  * @param {function} callback The callback to call with the authorized client.
  */
 export async function authorize(openurl?: OpenUrl) {
-  const credentialsPath = process.env.GMAIL_CREDENTIALS_PATH;
-  if (!credentialsPath || !existsSync(credentialsPath))
-    throw new Error("Cannot authorize gmail without credentials");
-  const credentials = JSON.parse(readFileSync(credentialsPath, "utf8"));
-  const { client_secret, client_id, redirect_uris } = credentials.installed
   const oAuth2Client = new google.auth.OAuth2(
```
FrozenKiwi, line 24: when this function is hit by the testing code, it doesn't matter what the credentials are because the google apis are mocked.  However, the test environment doesn't have the GMAIL_CREDENTIALS, so would throw for no good reason

libs/tx-gmail/src/index.live.test.ts
```diff
@@ -13,13 +15,16 @@ describe("Live service queries for gmail", () => {
   beforeAll(async () => {
     const timeout = 30 * 60 * 1000;
     jest.setTimeout(timeout);
-    ConfigStore.initialize();
-    await init('broker-cad');
+
+    // Initialize config with production data
+    const prodEnv = require('dotenv').config({path: getEnvFile("prod")})
```
FrozenKiwi, line 20: this only executes on manual runs, so is allowed to pull in prod vars

MarieTheCoin (approved):

PR #238, opened 2021-06-29
All/mocks

FrozenKiwi:
Centralized __mocks__ folder allows sharing mocks amongst all classes.

It also enables automock'ing for all inter-library calls within our packages (as is good practice - ideally, we limit the scope of the code under test to just the local library).

This PR is PoC, and does not complete the work of moving all the mocks into the shared space.  That will come after refactoring MarketStatus to allow mocking


apps/admin/jest.config.js
```diff
@@ -2,5 +2,9 @@ const defaults = require('../../jest.config');
 
 module.exports = {
   ...defaults,
-  moduleDirectories: ['node_modules', '../../node_modules', 'app']
+  moduleDirectories: ['node_modules', '../../node_modules', 'app'],
+  roots: [
+    defaults.roots[0],
+    '<rootDir>/app',
```
FrozenKiwi, line 8: The only project in our repo that uses 'app' instead of 'src' as it's root.

apps/broker-service/src/newsletter/newsletter.test.ts
```diff
@@ -1,15 +1,10 @@
-jest.mock('@thecointech/email');
```
FrozenKiwi, line 1: no need to remember to mock 'email' explicitly

apps/tx-processor/src/deposits/index.test.ts
```diff
@@ -5,7 +5,6 @@ import { getCurrentState } from '@thecointech/tx-processing';
 import { RbcApi } from '@thecointech/rbcapi';
 import { log } from '@thecointech/logging';
 
-jest.mock('@thecointech/email');
 jest.mock('@thecointech/utilities/MarketStatus', () => ({
```
FrozenKiwi, line 8: Mocking this next

jest.config.js
```diff
@@ -1,6 +1,14 @@
...
 const { compilerOptions } = require('./tsconfig.base.json');
+const { cwd } = require('process');
+
 const getTool = (name) => path.join(__dirname, 'tools', name);
+const getRoots = () => [
+  path.join(__dirname, '__mocks__'),
+  cwd().endsWith('TheCoin')
+    ? './'
+    : '<rootDir>/src'
+]
```
FrozenKiwi, line 11: As jest may either be run from root (via test explorer), or from an individual package, this jest config needs to adapt to either scenario with appropriate paths etc.

jest.config.js
```diff
@@ -10,24 +18,32 @@ module.exports = {
...
+        // typeRoots needs to be set absolutely because we lose the relative import of tsconfig.base.json
         typeRoots: [path.join(__dirname, "node_modules", "@types")],
-        incremental: true,
-        tsBuildInfoFile: '.tsbuildinfo',
+        // Disable incremental (as our tests aren't built anyway)
+        incremental: false,
+        // Point rootDir to root to allow compiling (__mocks__ folder)
+        rootDir: __dirname,
+        // Add multiple rootDirs
+        rootDirs: getRoots(),
```
FrozenKiwi: this isn't necessary (false lead) removing now

libs/tx-gmail/src/convert.test.ts
```diff
@@ -1,6 +1,7 @@
 import { toDepositData } from "./convert";
-import emails from './__mocks__/emails.get.json';
 import { log } from '@thecointech/logging';
+//@ts-ignore Import directly from mocks to skip auth part.
+import emails from '../../../__mocks__/emails.get.json';
```
FrozenKiwi, line 4: icky, but... /shrugs

PR #240, opened 2021-06-30
All/mocks2

FrozenKiwi:
Auto-mocking dev build for sites


.gitignore
```diff
@@ -62,9 +62,7 @@ libs/nft-contract/migrations/types
 libs/nft-contract/src/types
 libs/nft-contract/devlive.tokens.json
 
-libs/shared/src/containers/IDX/config.json
-libs/shared/src/containers/IDX/config.devlive.json
-
+libs/idx/src/config.devlive.json
```
MarieTheCoin, line 65: It would be good to find a way to share those files or at least the format. That one is not a problem but others might.

FrozenKiwi: The files are auto-generated - developers should never need to know they exist

We do commit the ones created for prod/prodtest though (they need to be persisted)

PR #246, opened 2021-07-06
fixed Coin update error when start date is OOR

FrozenKiwi:
On initialization, if no coin rate exists, the DB would fail to update.

This bug only affected prodtest, but theoretically could impact release if (for example) google went down for 3+ days


__mocks__/axios/index.ts
```diff
@@ -0,0 +1,21 @@
+import stock from './fetch.test.stock.json';
+import fx from './fetch.test.fx.json';
+
+// NOTE: This axios mock has not been used/tested outside of FinnHub/index.test.ts
```
FrozenKiwi, line 4: The mock turned out to not be useful in tracking this bug, but it is nice to have so left it in here

apps/rates-service/src/controllers/update.controller.ts
```diff
@@ -19,4 +20,13 @@ export class UpdateController extends Controller {
           throw e;
       }
     }
+
+    /**
+     * Test exception emailing
+     */
+    @Get('throwError')
+    @Response('500', 'always throws')
+    async throwError() {
```
FrozenKiwi, line 29: will remove this function later (after testing on deployed servers)

apps/rates-service/src/internals/rates/fetchCoin.ts
```diff
@@ -3,13 +3,19 @@ import { RateOffsetFromMarket, CoinUpdateInterval, CoinRate } from "./types";
...
   // This is to ensure we include the candle starting 1.5 mins
   // before our last validity expired
-  const fetchTimestamp = latestUntil - 6.5 * (60 * 1000);
-  return fetchNewCoinRates("1", fetchTimestamp, Date.now());
+  const now = DateTime.now();
+  // We can only fetch
+  let fetchTimestamp = Math.max(
+    latestUntil - 6.5 * (60 * 1000),
+    now.minus({days: 3}).toMillis()
+  );
```
FrozenKiwi, line 17: We seem to be limited to 3 days of history.

FrozenKiwi: (for 1-min results anyway)

MarieTheCoin (approved):

PR #247, opened 2021-07-06
Draft: Remove ids from messages - libs/shared and libs/site-base with ttypescript not working

MarieTheCoin:


.gitignore
```diff
@@ -64,7 +64,11 @@ libs/nft-contract/devlive.tokens.json
 
 libs/idx/src/config.devlive.json
 
-apps/site-nft/src/translations
-apps/site-nft/lang
+apps/site-nft/src/translations/fr.json
+apps/site-nft/src/translations/en.json
```
MarieTheCoin, line 68: Those should be build / compiled and not in git

PR #248, opened 2021-07-07
Broker/fix g drive storage

FrozenKiwi:
cleans up GDrive  API access to allow each configuration to have their own.

Currently implemented for prod:test


apps/site-landing/src/containers/Privacy.tsx
```diff
@@ -0,0 +1,70 @@
+import React from 'react';
+
+export const Privacy = () => (
+  <div>
+    <br />Privacy Policy
```
MarieTheCoin, line 5: Can / Should we make that translatable? It should probably be in every site

apps/site-landing/src/containers/TOS.tsx
```diff
@@ -0,0 +1,43 @@
+import React from 'react';
+
+export const TOS = () =>
+  <div>
+<br />Welcome to TheCoin Collaborative Canada
```
MarieTheCoin, line 5: Can / Should we make that translatable? It should probably be in every site

MarieTheCoin (approved):

apps/site-landing/src/containers/Privacy.tsx
FrozenKiwi: I wondered about that.  It's not really what I wanted to focus on with this PR though - I only need to put them in to pass google's site verification policy.

As for one in each site - I wondered about that.  I would lean towards just have one and each subdomain just links back to this one.

I'm not sure I would keep either of these though - they are probably insufficient for our special use-cases.  At some point, I think I will have to purchase something a bit more robust for TOS at the very least.

I'm not sure how many websites translate their TOS though

FrozenKiwi: Actually, we'll probably just borrow wealthsimples ones :-). But that is later-me's problem

PR #257, opened 2021-07-21
Translate site-landing

MarieTheCoin:


apps/site-landing/lang/en.json
```diff
@@ -1,428 +1,428 @@
 {
-  "'+yxUhg'": {
```
MarieTheCoin, line 2: Not sure where those "'" are coming from.

FrozenKiwi: I think I found them.  There were quotes in the package.json (check last commit) that seem to be treated differently in win/iOS

FrozenKiwi (approved):

PR #261, opened 2021-07-30
Admin/update

FrozenKiwi:
A massive trove of changes.

Admin app was based on a custom template from way-back.  Instead of attempting to update it, this PR simply removes the original app and builds a new one using `electron-forge`

There is also a fairly large grab-back of changes in here:

0) Admin app completely re-built
1) AuthRoute moved to shared, Routing has been standardized around the simpler concept trialed in site-nft
2) The Saga/Reducer code has finally been completed, with (nearly) all boilerplate pushed into a single set of base classes.  This change also fixes the typing issues.
3) ...


__mocks__/@thecointech/broker-cad/ETransferApi.ts
```diff
@@ -0,0 +1,21 @@
...
+    return buildResponse({
+      message: "Success",
+      state: "uh?",
+      hash: xfer.hash,
+    })
+  }
+  async eTransferInCode(request) {
+    await delay(250);
+    return buildResponse({ code: "A1B2C3" })
+  }
```
FrozenKiwi, line 20: new mocks - woo

__mocks__/mock_node.js
```diff
@@ -33,39 +21,48 @@ function mockAll() {
...
+      extensions: ['.js', '.json', '.node', '.ts', '.tsx'],
+      modules: [__dirname, 'node_modules'],
+    })
+
+  // NOTE: If we redirect all resolutions, breakpoints stop working
+  // Just redirect my own packages (and whichever external API's
+  // I want to have mocked) and everything works correctly
+  const toResolve = ['@thecointech', ...modules]
+  const oldResolve = Module._resolveFilename
+  Module._resolveFilename = (request, parent, isMain) => {
```
FrozenKiwi, line 37: Some serious hacking of node's internals.  Here we complete the port to conditional exports, by using the enhanved-resolver Node can use conditional exports (like webpack & jest)

__mocks__/prismic-javascript.ts
```diff
@@ -0,0 +1,21 @@
...
+      results_per_page: 10,
+      results_size: 0,
+      total_results_size: 0,
+      total_pages: 0,
+      next_page: '',
+      prev_page: '',
+      results: [],
+    }
+   }
+}
```
FrozenKiwi, line 17: Mocked prismic, needs data

apps/admin/src/containers/App/Routes.tsx
```diff
@@ -0,0 +1,16 @@
+import * as React from 'react';
+import { AuthSwitch } from '@thecointech/shared/containers/AuthRoute';
+import { NotFoundPage } from '@thecointech/shared/containers/NotFoundPage';
+import { AccountId } from '@thecointech/signers';
+import { BrokerCAD } from '../BrokerCAD';
+import { TheCoin } from '../TheCoinAccount';
+
+export const AppRoutes = {
```
FrozenKiwi, line 8: porting the new routing layout to admin

apps/admin/src/containers/Purchase/index.tsx
```diff
@@ -0,0 +1,88 @@
...
+  const [step, setStep] = React.useState("");
+
+  const account = AccountMap.useActive()!;
+  const api = Account(account.address).useApi();
+
+  const onConfirm = async () => {
+    setConfirm(false);
+    setIsProcessing(true);
+    setForceValidate(true);
+    throw new Error("Just because it compiles, doesn't mean it works.  Check properly");
```
FrozenKiwi, line 38: This class is for manual purchases.  It probably won't ever gt used again

apps/admin/src/init/sidebar.ts
```diff
@@ -0,0 +1,47 @@
...
+  purchase: { defaultMessage: "Complete Purchase", description: "Title for the purchase entry in the menu" },
+
+  eTransfer: { defaultMessage: "Complete e-Transfer", description: "Title for the Complete e-Transfer entry in the menu" },
+  billing: { defaultMessage: "Bill Payments", description: "Title for the Bill Payments entry in the menu" },
+  verify: { defaultMessage: "Verify", description: "Title for the Verify entry in the menu" },
+  autoPurchase: { defaultMessage: "AutoPurchase", description: "Title for the AutoPurchase entry in the menu" },
+  clients: { defaultMessage: "Clients", description: "Title for the Clients entry in the menu" },
+});
+
+const buildSubMenu = (id: AccountId, routes: typeof TheCoinRoutes|typeof BrokerCADRoutes) =>
```
FrozenKiwi, line 22: The construction of the sidebar has been simplified considerably

apps/admin/webpack.mocked.config.js
```diff
@@ -0,0 +1,31 @@
...
+          },
+        },
+      }
+    ],
+  },
+  // Re-use our jest mocks inside our website (neat, huh?)
+  resolve: {
+    conditionNames: [process.env.CONFIG_NAME, "electron", "browser", "require", "default"],
+    modules: [mocksFolder]
+  }
```
FrozenKiwi, line 30: This is obvs not ready for production yet

apps/admin/webpack.plugins.js
```diff
@@ -0,0 +1,23 @@
...
+    __VERSION__: JSON.stringify(require(packageFile).version),
+    "process.env.LOG_NAME": JSON.stringify(process.env.LOG_NAME),
+    "process.env.LOG_LEVEL": process.env.LOG_LEVEL,
+  }),
+  new webpack.ProvidePlugin({
+    process: 'process/browser',
+  }),
+  new webpack.ProvidePlugin({
+    Buffer: ['buffer', 'Buffer'],
+  }),
```
FrozenKiwi, line 22: These webpack configs could be shared from websites

libs/contract/truffle-config.js
```diff
@@ -1,15 +1,17 @@
...
+// Allow using typescript in deployments
+loadTypescript();
+
+var path = require('path');
+const HDWalletProvider = require("@truffle/hdwallet-provider");
+const { AccountId, getSigner } = require('@thecointech/signers');
 
+const numBuiltIn = AccountId.BrokerCAD + 1;
+const testAccounts = [];
+loadAccounts(numBuiltIn).then(v => testAccounts = v).catch(console.error);
```
FrozenKiwi, line 14: slightly cheat-y and quite risky.  No top-level async here, and we can no longer access signers directly (they are env-specific now).  So, just assume that we will load signers before the code finishes compiling

libs/shared/src/components/CopyToClipboard/index.tsx
```diff
@@ -28,6 +31,8 @@ export const CopyToClipboard: FunctionComponent<Props> = (props) => {
 
   const onClick = (event: React.MouseEvent<HTMLAnchorElement, MouseEvent>) => {
     Clipboard.copy(props.payload);
+    setOpen(true);
+    setTimeout(() => setOpen(false), 7000);
     event.stopPropagation();
```
FrozenKiwi, line 36: stories are awesome for catching little UX issues like this one (notificiation did not show if clicking on text tobe copied)

libs/shared/src/containers/PageSidebar/index.tsx
```diff
@@ -1,55 +1,71 @@
...
+const MenuItems = ({generators, items}: SidebarState) => {
+  let final = Object
+    .values(generators) // If we ever go back to multiple generators, we should sort by priority
+    .reduce((prev, generator) => generator(prev), items);
 
-  return buildMenuArray(items);
+  return (
+    <>
+      {final.header ? <SidebarHeader {...final.header} /> : null}
+      <MenuLinks links={final.links} />
```
FrozenKiwi, line 45: There is no reason to merge header & links

libs/shared/src/store/immerReducer.ts
```diff
@@ -1,68 +1,180 @@
...
+import { ImmerReducer, createReducerFunction, createActionCreators, ImmerReducerClass, ImmerReducerState } from "immer-reducer";
+import { put, select, SelectEffect, takeLatest } from "redux-saga/effects";
+import { Action, bindActionCreators, Reducer } from "redux";
+import { useDispatch, useSelector } from 'react-redux';
+import { useInjectReducer, useInjectSaga } from 'redux-injectors';
+import { Saga } from 'redux-saga';
 
-interface Action {
-  type: string;
+
```
FrozenKiwi, line 8: This file is particularily obtuse, but I'm also particularily proud of it!

FrozenKiwi: Bascially, it hides all the pain of redux & sagas in one place

libs/shared/src/store/immerReducer.ts
```diff
@@ -1,68 +1,180 @@
...
+  [K in keyof Interface] : ImmerActionCreator<K, ArgumentsType<Interface[K]>>
 }
 
-export class TheCoinReducer<T> extends ImmerReducer<T> {
+export function BaseReducer<U, T>(key: string, initialState: T) {
+  // Return new class so we can re-use the static declarations
+  const BaseReducer = class BaseReducer extends ImmerReducer<T> {
+
+    static _reducer: any;
+    static _actions: any;
```
FrozenKiwi, line 32: yes, `any` - but the properly-typed vesrions are below, this is just to get around the fact that we don't actually have proper definitions for these types yet

__mocks__/mock_node.js
MarieTheCoin: A bit scary

apps/admin/config.forge.js
```diff
@@ -0,0 +1,44 @@
+module.exports = {
+  packagerConfig: {},
+  makers: [
+    {
+      name: "@electron-forge/maker-squirrel",
```
MarieTheCoin, line 5: What is this file for?

apps/admin/src/containers/App/Routes.tsx
MarieTheCoin: New routing layout? Is it something we will have to do everywhere?

apps/admin/src/containers/FirestoreSignin/firestore.ts
```diff
@@ -16,7 +16,7 @@ export async function signIn() {
     if (!username || !password)
         return false;
 
-    return await init({username, password});
+    return init({username, password});
```
MarieTheCoin, line 19: No await necessary?

apps/admin/src/containers/Purchase/index.tsx
MarieTheCoin: Should it be named "obsolete" of "legacy"? 😄

MarieTheCoin: Joke aside we could put a log line in here and check in a while if it's used at all.

apps/site-app/src/containers/Sidebar/index.tsx
```diff
@@ -0,0 +1,128 @@
...
+  const intl = useIntl();
+
+
+  // Update the header whenever the active account is changed.
+  React.useEffect(() => {
+    sidebarApi.setHeader({
+      title: translations.profile,
+      avatar: getAvatarLink("14"),
+      accountName: active?.name ?? intl.formatMessage(translations.noAccount),
+      address: active?.address ?? "-----------",
```
MarieTheCoin, line 106: Is this right? The "-----------"

apps/site-app/src/translations/en.json
```diff
@@ -41,7 +41,6 @@
   "'InBntU'": "Settings",
   "'It9lF6'": "Settings",
   "'JHN0Qb'": "Transfer funds from your Canadian bank to your account.",
-  "'JgdFdU'": "Profile",
   "'K7rhf2'": "Settings",
```
MarieTheCoin, line 44: We should not commit those they should be compiled when build. It could become a conflict problem and it pollutes git.

libs/shared/src/components/CopyToClipboard/index.tsx
```diff
@@ -9,15 +9,18 @@ interface Props {
...
-      id: "base.copyToClipboard.link",
-      defaultMessage:"Copy",
-      description:"base.copyToClipboard.link: Link for the button for the copy to clipboard tool"},
-    messageSuccess : { 
-      id: "base.copyToClipboard.messageSuccess",
-      defaultMessage:"Copied to your clipboard",
-      description:"base.copyToClipboard.messageSuccess: Success message for the button for the copy to clipboard tool"}});
+const translate = defineMessages({
+  link: {
+    id: "base.copyToClipboard.link",
```
MarieTheCoin, line 14: Is it possible to remove those ids?

libs/shared/tsconfig.json
```diff
@@ -4,15 +4,7 @@
...
     "outDir": "./build/",
-    "tsBuildInfoFile": "./build/.buildinfo",
-    "plugins": [
-      {
-        "transform": "@formatjs/ts-transformer",
-        "import": "transform",
-        "type": "config",
-        "overrideIdFn": "[sha512:contenthash:base64:6]"
-      }
-    ]
```
MarieTheCoin, line 15: This was added in admin and removed here. Is it normal?

MarieTheCoin (approved):

apps/admin/src/containers/Purchase/index.tsx
FrozenKiwi: It runs in Admin.  The only person who see's it is me (for the foreseeable future).

After thinking about it, it will probably be resurrected at some point for non-ETransfer based transfers.  Which will (hopefully) happen

libs/contract/truffle-config.js
FrozenKiwi: side note: it's risky in that it might not work in production, but not risky in that if it does break (we'll have to see when this code gets to that point) it will be extremely obvious and nothing in the system will break

apps/admin/config.forge.js
FrozenKiwi: Yeah, `Electron-Forge` is the package that runs our app.  It's kinda like `create-react-app`, just for electron.

apps/admin/src/containers/FirestoreSignin/firestore.ts
FrozenKiwi: nope, turns out it was necessary, I've put it back in in the next PR (d'oh)

apps/site-app/src/containers/Sidebar/index.tsx
FrozenKiwi: well... I mean... I'm not entirely sure what we could show instead...

The main thing is that the sidebar shouldn't show if we don't have an account loaded.  So it shouldn't ever actually trigger that code.  Maybe in some future world that changes though, so the line should let whoever needs to know that that it's breaking things

apps/site-app/src/translations/en.json
FrozenKiwi: yeah, it's a real pain.  Also guaranteed conflicts when anything changes.  As I'm going through the translation PR's I'm removing these.

libs/shared/src/components/CopyToClipboard/index.tsx
FrozenKiwi: Eeek - ah, nuts.   I see that's in your site PR that I'm working on now.  Hopefully git can figure out the merge.  I'll merge that PR into here as soon as it's complete (I added a check to make sure no committed files are changed on build - basically to check that any translation changes are managed on developer end.  However, it seems to be failing on unrelated issues, and I might have to remove it)

libs/shared/tsconfig.json
FrozenKiwi: Yeah, that's consistent with the rest of the projects... (I think).

It kinda makes no difference - VSCode-version TS will ignore it, and the build will get it either way.  But because VS-Code version of TS ignores these instructions, it seems tidier to have them just in the build file so it's clear where they are applied.

apps/admin/webpack.mocked.config.js
FrozenKiwi: I'm working on that now.  Just trying to figure out exactly how handle the switching.

We now have webpack/node/jest & this add's Electron into our build config shenanigans.  It's not super-fun to balance them all.

apps/admin/src/containers/Purchase/index.tsx
FrozenKiwi: But that's a good point: better than just 'throwing' (which is quite unexpected).  I'll disable the button & change the text to make it clear this component is unusable

apps/admin/src/containers/App/Routes.tsx
FrozenKiwi: I tried to do it everywhere in this PR.  It's part of why it got so humungous.  (but note: 5000 LoC dropped - even most of them were in `yarn.lock`!

__mocks__/mock_node.js
FrozenKiwi: if it works once, it will work forever!!!

PR #263, opened 2021-07-31
Fix restore 'goto' step in loading saved accounts

FrozenKiwi:
Initial branch for feature "Restore Accounts"

The functionality is already more-or-less complete.  However past that the page looks pretty pants.

image

The full page running in context looks like this:

https://user-images.githubusercontent.com/4876160/127738916-9795b280-9aff-4806-9eae-a6cb4cb27ed7.mp4


MarieTheCoin (approved):
I worked on the style in another PR: https://github.com/thecointech/thecoin/pull/264

PR #266, opened 2021-08-04
All/refactor tx processor

FrozenKiwi:
We will need to do a bit of work (including introducing a dependency on electron) to get tx-gmail running cleanly in electron.

This would have a large follow-on effect - we don't want electron being installed in our web servers.

We break tx-processor into tx-deposit/tx-etransfer/tx-bills instead.  tx-deposit only runs on the local machines so we simply don't include it on our web services


FrozenKiwi:
The majority of this PR is just copy/paste/rename.  No functionality was harmed in the making of this PR

libs/tx-statemachine/src/transitions/index.ts
```diff
@@ -1,13 +1,11 @@
-export * from './depositFiat';
 export * from './depositCoin';
 export * from './pretransfer';
 export * from './sendCoin';
 export * from './toCoin';
 export * from './waitCoin';
 export * from './timeout';
 export * from './noop';
```
MarieTheCoin, line 7: noop: my favorite :)

MarieTheCoin (approved):

libs/tx-statemachine/src/transitions/index.ts
FrozenKiwi: he works as hard as I do ;-)

PR #267, opened 2021-08-06
Fix create account process web3

MarieTheCoin:
Capture d’écran, le 2021-08-06 à 18 28 16


libs/site-base/package.json
```diff
@@ -25,7 +25,7 @@
     "test": "cross-env NODE_ENV=test jest --coverage",
     "test:watch": "cross-env NODE_ENV=test jest --watchAll",
     "coveralls": "cat ./coverage/lcov.info | coveralls",
-    "extract": "formatjs extract \"src/**/!(*.d|*.test).ts*\" --out-file lang/en.json --id-interpolation-pattern [sha512:contenthash:base64:6]c",
```
MarieTheCoin, line 28: I was crushed when I saw all the ids for the translations changed after a yarn extract. 
Then I felt relieved to see that :)

libs/site-base/src/containers/AddAccount/CreateExistingSwitch/index.tsx
```diff
@@ -47,13 +42,23 @@ export const CreateExistingSwitch = (props: Props) => {
...
+        <div className={`x8spaceBefore`}>
+            <Link className={styles.linkEthereum} to={`${url}connect/create`}>
+              <FormattedMessage {...translations.ethereumNewAccount} />
+            </Link>
+        </div>
+        <div>
+            <Link className={styles.linkEthereum} to={`${url}connect/exist`}>
+              <FormattedMessage {...translations.ethereumExistingAccount} />
+            </Link>
+        </div>
```
MarieTheCoin, line 61: Wasn't sure if both links are needed. Tell me if you want me to remove one of them.

libs/site-base/package.json
FrozenKiwi: haha - yeah, we've wasted enough work already to lose any more!

libs/site-base/src/containers/AddAccount/CreateExistingSwitch/index.tsx
FrozenKiwi: We'll leave both in for now.  We can wait until we see user behaviour to see if we want to reduce it at all

libs/site-base/package.json
FrozenKiwi: also... woops!!! How on earth did that sneak in there?  Good catch!

FrozenKiwi (approved):
Your a gem Marie!  Thanks so much!

PR #269, opened 2021-08-09
Admin/disable node intergration

FrozenKiwi:
Our admin app previously allowed nodejs integration with the rendering process.

This is pretty bad for security, so this PR removes those privledges.  The major speed-bump was tx-gmail, which uses the nodejs API to fetch deposit e-transfers for emails.  This PR refactors that project so with an electron-specific version which uses IPC to communicate with itself on the node backend.


apps/admin/src/index.ts
```diff
@@ -18,11 +21,15 @@ const createWindow = (): void => {
...
+
     webPreferences:
     {
-      nodeIntegration: true,
-      contextIsolation: false,
-      nativeWindowOpen: true
+      nodeIntegration: false,
+      enableRemoteModule: false,
+      contextIsolation: true,
+      sandbox: true,
```
FrozenKiwi, line 30: security... now on...

apps/admin/src/init/gmail.ts
```diff
@@ -0,0 +1,17 @@
...
+
+declare let window: Window & {
+  ipcRenderer: Pick<IpcRenderer, "invoke">
+};
+
+export async function initGmail() {
+  // open IPC bridge to node process.
+  gmail.bridge(window.ipcRenderer);
+  // Cycle the gmail token
+  const token = await ConfigStore.get("gmailcred")
```
FrozenKiwi, line 14: we store the token in the renderer thread to keep it within ConfigStore.
It just means one less file written to disk (this db is stored in electrons localStorage)

apps/admin/src/preload.ts
```diff
@@ -0,0 +1,16 @@
+import { contextBridge, ipcRenderer } from "electron";
+// import gmail from '@thecointech/tx-gmail';
+
+// Whitelist limits invokes to tx-gmail
```
FrozenKiwi, line 4: whitelist unfortunately broke the compiler, I should remove this code.

The whitelist probably isn't necessary, as invoke only works if there is an existing "handle" call on the node side anyway

MarieTheCoin (approved):

PR #270, opened 2021-08-15
Shop Database JPG

MarieTheCoin:
shop-database


FrozenKiwi:
This is good - the only thing we are missing is discounts (the plan is to give existing users discount codes based on their TC account balance).

I will be working on the sales immediately after posting the NFT squeeze I'll probably start on this next week

FrozenKiwi:
I'll merge this in, because it seems workable as-is, and we can discuss discounts/credit later.

PR #273, opened 2021-08-24
Nft/subscription

FrozenKiwi:
Include subscription widget
Convert SVG to PNG/WebP
Tweak CSS so we don't lose sub widget off the end of the page

Also, added a bunch of text so I can at least post this and it's not quite so WIP

image

image


FrozenKiwi:
OK - I think it's actually finished now.

apps/nft-squeeze/buildimgs.ts
```diff
@@ -0,0 +1,22 @@
+import * as glob from 'glob';
+import * as sharp from 'sharp'
+
+async function convertFiles() {
+  const files = glob.sync("./src/images/*.svg");
```
FrozenKiwi, line 5: This is an experiment - the build step converts from SVG into png/webp.  It is something we may consider for the other sites to improve load times

MarieTheCoin (changes requested):

PR #278, opened 2021-08-30
Fix/load accounts

FrozenKiwi:
Account storage was not properly loaded from the 'accounts' module.

The core problem was top-level async did not propagate through the libraries.

### DIFF
--- a/libs/account/src/devlive/store.ts
+++ b/libs/account/src/devlive/store.ts
@@ -2,8 +2,9 @@ import { ConnectContract } from '@thecointech/contract';
 import { getSigner, AccountName } from '@thecointech/signers';
 import { NormalizeAddress } from '@thecointech/utilities/Address';
 import { AccountState, buildNewAccount } from '../state';
+import * as Browser from '../store';
 
-const _devWallets = {} as Record<string, AccountState>;
+const _devWallets = Browser.getAllAccounts();
 let _initial = null as string|null;
 
 async function addRemoteAccount(name: AccountName, active: boolean) {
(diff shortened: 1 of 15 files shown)


libs/account/src/devlive/store.ts
```diff
@@ -2,8 +2,9 @@ import { ConnectContract } from '@thecointech/contract';
 import { getSigner, AccountName } from '@thecointech/signers';
 import { NormalizeAddress } from '@thecointech/utilities/Address';
 import { AccountState, buildNewAccount } from '../state';
+import * as Browser from '../store';
 
-const _devWallets = {} as Record<string, AccountState>;
+const _devWallets = Browser.getAllAccounts();
```
FrozenKiwi, line 7: include previously saved accounts to make dev:live better testing environment.

`dev` resets all accounts on every run, but dev:live should be close to final environment

libs/account/src/store.ts
```diff
@@ -67,7 +67,7 @@ export function getAllAccounts(): AccountMap {
       const { address, signer } = account;
 
       // Rough check that this is, indeed, a valid account
-      if (address && IsValidAddress(address) && signer._isSigner !== undefined) {
```
FrozenKiwi, line 70: Upgrade to ethers5 nixed this symbol

libs/contract/truffle-config.js
```diff
@@ -11,7 +11,10 @@ const { AccountId, getSigner } = require('@thecointech/signers');
 
 const numBuiltIn = AccountId.BrokerCAD + 1;
 const testAccounts = [];
-loadAccounts(numBuiltIn).then(v => testAccounts.push(...v)).catch(console.error);
+if (process.env.CONFIG_NAME !== 'devlive') {
+  // devlive accounts are hosted on our local blockchain, so already available
+  loadAccounts(numBuiltIn).then(v => testAccounts.push(...v)).catch(console.error);
+}
```
FrozenKiwi, line 17: This line would throw an error because the signers loaded are not wallets (remote signers from blockchain, and this code is loading them back onto the blockchain.  Very circular, and unnecessary)

libs/site-base/internal/webpack/webpack.prod.babel.js
```diff
@@ -98,6 +98,9 @@ const prodOptions = {
     assetFilter: assetFilename =>
       !/(\.map$)|(^(main\.|favicon\.))/.test(assetFilename),
   },
+  experiments: {
+    topLevelAwait: true,
+  }
 };
```
FrozenKiwi, line 104: Isn't currently used on prod (only TLA is in dev:live specific code) but better to be consistent

libs/site-base/src/containers/AddAccount/FromJson/index.tsx
```diff
@@ -29,17 +29,14 @@ function readFile(file: File): Promise<ReadFileData> {
   });
 }
 
-export const FromJson = () => {
-  return (
-    <Container>
-      <UploadWallet readFile={readFile}/>
-      <div>
-          <FormattedMessage {...translations.explanation} />
+export const FromJson = () =>
```
FrozenKiwi, line 32: NO CHANGE:

Needed an update to trigger rebuild is all

PR #279, opened 2021-08-31
Fix/e transfer page

FrozenKiwi:
fix some superficial issues with the eTransfer page

### DIFF
--- a/apps/admin/src/containers/Purchase/index.tsx
+++ b/apps/admin/src/containers/Purchase/index.tsx
@@ -12,6 +12,7 @@ import messages from './messages';
 import { getCurrentState } from '@thecointech/tx-statemachine';
 import { AccountMap } from '@thecointech/shared/containers/AccountMap';
 import { Account } from '@thecointech/shared/containers/Account';
+
 const typeOptions = ["deposit", "other"].map(k => ({
   key: k,
   value: k,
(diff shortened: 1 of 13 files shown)


apps/admin/src/containers/Purchase/index.tsx
```diff
@@ -72,7 +71,7 @@ export const Purchase = () => {
         <Form.Button onClick={() => setConfirm(true)}>SEND</Form.Button>
       </Form>
       <Confirm open={confirm} onCancel={() => setConfirm(false)} onConfirm={onConfirm} />
-      <ModalOperation isOpen={isProcessing} header={messages.mintingHeader} progressMessage={messages.mintingInProgress} messageValues={{ step }} />
+      <ModalOperation isOpen={isProcessing} header={messages.mintingHeader} progressMessage={messages.mintingInProgress} />
```
FrozenKiwi, line 74: Passing the values through as a separate entry could cause issues when the component re-rendered too quickly (ie after setting message, but before value, or vice-versa).

By merging message & value, we eliminate the possibility that the we might render during a transition state

apps/site-app/src/containers/Accounts/BillPayments/payees.ts
```diff
@@ -19,7 +19,7 @@ const translations = defineMessages({
 });
 
 export type Validatable = {
-  validate: (value: string) => ValuedMessageDesc|null,
```
FrozenKiwi, line 22: Also had a type define with the same purpose, so merged them together for both UxInput & ModalOperation

PR #280, opened 2021-08-31
Fix/history repeat load

FrozenKiwi:
When navigating pages, the history pane will update transactions.  The code would re-calculate all balances, but this modified state in redux, (is disallowed) which then caused an exception.

### DIFF
--- a/libs/shared/src/containers/Account/reducer.ts
+++ b/libs/shared/src/containers/Account/reducer.ts
@@ -7,7 +7,7 @@ import { buildSagas } from './actions';
 import { FxRateReducer } from '../../containers/FxRate/reducer';
 import { SagaReducer } from '../../store/immerReducer';
 import { isLocal } from '@thecointech/signers';
-import { loadAndMergeHistory, calculateTxBalances, Transaction } from '@thecointech/tx-blockchain';
+import { loadAndMergeHistory, calculateTxBalances, mergeTransactions, Transaction } from '@thecointech/tx-blockchain';
 import { connectIDX } from '@thecointech/idx';
 import { AccountDetails, AccountState, DefaultAccountValues } from '@thecointech/account';
 import { loadDetails, setDetails } from '../AccountDetails';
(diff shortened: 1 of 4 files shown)


libs/shared/src/containers/Account/reducer.ts
```diff
@@ -119,22 +119,18 @@ function AccountReducer(address: string, initialState: AccountState) {
...
-
-      const origHistory = this.state.history;
+      log.trace(`Updating from ${from} -> ${until}`);
+      const oldHistory = this.state.history;
       const fromBlock = this.state.historyEndBlock || InitialCoinBlock;
       // Retrieve transactions for all time
-      const newHistory: Transaction[] = yield call(loadAndMergeHistory, address, fromBlock, contract, origHistory);
+      const newHistory: Transaction[] = yield call(loadAndMergeHistory, address, fromBlock, contract);
       // Take current balance and use it plus Tx to reconstruct historical balances.
       calculateTxBalances(balance, newHistory);
```
FrozenKiwi, line 128: fix: only recalculate balances on the newly fetched history

libs/tx-blockchain/src/fetch.ts
```diff
@@ -102,66 +88,20 @@ async function readAndMergeTransfers(account: string, to: boolean, fromBlock: nu
...
-//   }
-//   // merge and remove duplicates for complete array
-//   return mergeTransactions(history, txs);
-// }
-
-export async function loadAndMergeHistory(address: string, fromBlock: number, contract: TheCoin, history: Transaction[]) {
+export async function loadAndMergeHistory(address: string, fromBlock: number, contract: TheCoin) {
   try {
-    //history = await readAndMergePurchases(address, fromBlock, contract, history);
+    let history: Transaction[] = [];
```
FrozenKiwi, line 93: This is a slight behaviour change.  When reading transactions, we may get duplicates when reading to & from.  However, duplicates happen within the same block, so while we may find duplicates in the merge of from & to, we won't ever have duplicates between different runs of this function

FrozenKiwi: (assuming, of course, that fromBlock is correctly set)

libs/tx-blockchain/src/fetch.ts
```diff
@@ -102,66 +88,20 @@ async function readAndMergeTransfers(account: string, to: boolean, fromBlock: nu
...
     history = await readAndMergeTransfers(address, true, fromBlock, contract, history);
     history = await readAndMergeTransfers(address, false, fromBlock, contract, history);
+    return history;
   }
   catch (err) {
-    console.error(err);
+    log.error(err);
```
FrozenKiwi, line 99: Question: is it better to catch or let any exceptions propagate.

Our exception handling is not consistent

PR #282, opened 2021-08-31
Fix/null error on e transfer

FrozenKiwi:


apps/admin/src/containers/Purchase/index.tsx
```diff
@@ -72,7 +71,7 @@ export const Purchase = () => {
         <Form.Button onClick={() => setConfirm(true)}>SEND</Form.Button>
       </Form>
       <Confirm open={confirm} onCancel={() => setConfirm(false)} onConfirm={onConfirm} />
-      <ModalOperation isOpen={isProcessing} header={messages.mintingHeader} progressMessage={messages.mintingInProgress} messageValues={{ step }} />
+      <ModalOperation isOpen={isProcessing} header={messages.mintingHeader} progressMessage={messages.mintingInProgress} />
```
FrozenKiwi, line 74: combining values & message requires the client to ensure the data is consistent

apps/site-app/src/containers/Accounts/Redeem/index.tsx
```diff
@@ -137,31 +136,25 @@ export const Redeem = () => {
...
+          <a target="_blank" href={`https://ropsten.etherscan.io/tx/${response.data.hash}`}> here </a>),
+      }
+    });
     setPercentComplete(0.5);
-    setTransferValues(transferValues);
 
     const tx = await contract.provider.getTransaction(response.data.hash);
     // Wait at least 2 confirmations
     tx.wait(2);
-    const receipt = await contract.provider.getTransactionReceipt(
```
FrozenKiwi, line 151: not sure why this would return null.  However, this info doesn't really seem to have any use, so removed it

libs/shared/src/types/index.ts
```diff
@@ -34,3 +35,8 @@ export interface InjectSagaParams<S extends Saga> {
   saga: S;
   mode?: string | undefined;
 }
+
+// Allow passing values directly into MessageDescriptor
+export type MessageWithValues = {
+  values?: Record<string, any>
+} & MessageDescriptor;
```
FrozenKiwi, line 42: not a great place to put this.

PR #283, opened 2021-09-03
Feature/manual tx override

FrozenKiwi:
God-mode for the admin app.

This PR allows the admin app to manually reset the state of a transaction that has entered error mode.  This allows us to restart transactions that have failed for any reason, or to put them in refund/etc state.

This PR also more-or-less permanently disables the ability of the admin app to process transactions.  It is expected that all processing continue to happen in the processor, and the admin app only has the ability to change states.  It limits the information exposed in the admin app, and enforces the separation of concerns.


__mocks__/mock_node.js
```diff
@@ -6,36 +6,47 @@
 //
 
 const Module = require('module').Module;
+const ehr = require('enhanced-resolve');
+const fs = require('fs');
+const path = require('path');
+
+const allMocks = fs.readdirSync(__dirname).map(f => path.parse(f).name);
+
```
FrozenKiwi, line 15: I'm pretty sure I got it this time...

__mocks__/mock_node.js
```diff
@@ -53,7 +64,7 @@ switch (process.env.CONFIG_NAME) {
     // Dev live is internally connected, but all external connections are mocked
     //
     console.warn('--- Injecting external TC mocks ---');
-    useSomeMocks("googleapis", "google-auth-library");
+    useSomeMocks(["googleapis", "google-auth-library"]);
```
MarieTheCoin, line 67: Misread "useSomeSocks" and was a bit puzzled 😁

admin.code-workspace
```diff
@@ -2,39 +2,46 @@
...
-      "path": "libs\\logging"
+      "path": "libs/logging"
     },
     {
       "name": "contract",
-      "path": "libs\\contract"
+      "path": "libs/contract"
+    },
+    {
+      "path": "libs/signers"
```
MarieTheCoin, line 44: No name for this one?

libs/signers/src/fromDisk.ts
```diff
@@ -18,7 +18,7 @@ function loadEncrypted(name: AccountName) {
     if (path) {
       if (!existsSync(path))
         throw new Error(`Could not load ${name} from path: ${path}`);
-      wallet = readFileSync(path, 'utf8');
+      wallet = readFileSync(path, 'ascii');
```
MarieTheCoin, line 21: ascii instead of utf8?

libs/tx-statemachine/src/error.test.ts
```diff
@@ -0,0 +1,111 @@
...
+const transitionBase = (type: string) =>({
+  timestamp: DateTime.now(),
+  type,
+  // Reset errors/meta
+  error: undefined,
+  meta: undefined,
+});
+// Simple transitions just test the different kind of scenarios the FSM needs to process.
+// noop - no change to data, just transition to a new state
+const noop = async () => transitionBase('noop');
```
MarieTheCoin, line 22: noop

libs/tx-statemachine/src/error.test.ts
```diff
@@ -0,0 +1,111 @@
...
+  },
+  testError: {
+    next:  FSM.transitionTo<States>(noop, "complete"),
+    onError: FSM.transitionTo<States>(noop, "error"),
+  },
+  error: { next: manualOverride },
+  complete: null,
+}
+
+it("Error replay is handled appropriately", async () => {
```
MarieTheCoin, line 46: What is replay? The transaction?

MarieTheCoin (approved):

libs/signers/src/fromDisk.ts
FrozenKiwi: yeah, one of my accounts was loading with a weird leading character in it that broke JSON parsing.  I have no idea how it got in there, I assume that at some point the file was saved as Unicode and that was the leading byte for it.  Forcing it to `ASCII` strips the byte.   This kinda runs counter to most of what I thought I knew about unicode, but...

These are machine-files, it doesn't include any user-generated content, so the Utf8 is not necessary.

libs/tx-statemachine/src/error.test.ts
FrozenKiwi: A replay event is one that has occurred before.

Our transactions are restart-able.  This allows them to handle situations like the congested networks gracefully, they just time-out and then restart when things die down again.  When restarting, they replay existing events to get the transaction back into the same state it was in before the error.

libs/tx-statemachine/src/error.test.ts
FrozenKiwi: the only state change guaranteed not to crash

admin.code-workspace
FrozenKiwi: huh, weird.  It shows up with a name.  I've never explicitly named the other ones, so maybe this is a new feature in VS Code?

__mocks__/mock_node.js
FrozenKiwi: For the warm days.  Cold days it's `useAllSocks`!

libs/tx-statemachine/src/error.test.ts
FrozenKiwi: I've just realized that our docs don't actually cover this anywhere

PR #293, opened 2021-09-06
Fix storybook and add it to build step to ensure we don't lose it again

FrozenKiwi:
A recent PR broke storybook (https://github.com/thecointech/thecoin/pull/278).  This PR fixes storybook, as well as adds it to our CI build step to ensure we do not break it again in future.


package.json
```diff
@@ -22,9 +22,10 @@
     "gen:nft": "openapi-generator-cli generate -i apps/nft-service/src/api/swagger.json -g typescript-axios -o apis/nft/ts -c apis/nft/client-ts-spec.json",
     "genapis": "lerna run gen:swagger && cross-env JAVA_OPTS=\"-Dlog.level=warn\" run-p gen:*",
     "build:src": "lerna run build",
-    "build:all": "run-s clean genapis build:src",
-    "rebuild": "cross-env CONFIG_NAME=prod yarn build:all",
-    "build": "cross-env CONFIG_NAME=prod yarn build:src",
+    "build:storybook": "build-storybook",
+    "build:all": "run-s clean genapis build:src build:storybook",
+    "build": "cross-env CONFIG_NAME=prod yarn build:all",
+    "rebuild": "cross-env CONFIG_NAME=prod yarn build:src",
```
FrozenKiwi, line 28: NOTE: Changing build command behaviour.

Previously, `rebuild` was the complete rebuild, and `build` was incremental.  However, this is non-standard, and occasionally would result in errors when we used the incremental build when we needed full build.

FrozenKiwi: So we've reverted back to having the default `build` command clean & build everything, and `rebuild` is the one we can use if we explicitly want an incremental build

.storybook/main.js
```diff
@@ -57,6 +58,11 @@ module.exports = {
         }
       },
       config);
+    // Exclude our build folder from compilation.
+    // This is because shared/site-base build to ES2017
+    // which somehow breaks the FormatJS babel plugin.
+    const babelRule = r.module.rules.find(rule => rule.include?.includes?.(rootFolder));
+    babelRule.exclude = /(node_modules)|(build)/
```
FrozenKiwi, line 65: This is the build fix.  In our mono-repo, `@thecointech/shared/file` resolves to `/libs/shared/build/file`, and so would end up being re-processed by babel.

This isn't a big deal, but when we shifted `shared/site-base` to ESNext modules (to support root-level await), something broke in the FormatJS parser.  We don't -need- to parse these files, so easiest fix is to prevent Babel from doing that.

package.json
```diff
@@ -22,9 +22,10 @@
     "gen:nft": "openapi-generator-cli generate -i apps/nft-service/src/api/swagger.json -g typescript-axios -o apis/nft/ts -c apis/nft/client-ts-spec.json",
     "genapis": "lerna run gen:swagger && cross-env JAVA_OPTS=\"-Dlog.level=warn\" run-p gen:*",
     "build:src": "lerna run build",
-    "build:all": "run-s clean genapis build:src",
-    "rebuild": "cross-env CONFIG_NAME=prod yarn build:all",
-    "build": "cross-env CONFIG_NAME=prod yarn build:src",
+    "build:storybook": "build-storybook",
```
FrozenKiwi, line 25: no more broken storybook!

MarieTheCoin (approved):

PR #299, opened 2021-09-15
Fix/idx

FrozenKiwi:


libs/shared/src/containers/AccountDetails/persist.ts
```diff
@@ -1,8 +1,7 @@
-import { IDX } from "@ceramicstudio/idx";
-import { loadEncrypted, setEncrypted } from "@thecointech/idx";
+import { IDX, loadEncrypted, setEncrypted } from "@thecointech/idx";
 import { AccountDetails } from "@thecointech/account";
 
-export const setDetails = (idx: IDX, details: AccountDetails) =>
+export const setDetails = (idx: IDX, details: AccountDetails): Promise<any> =>
```
FrozenKiwi, line 4: this returns a type that is not defined here, and potentially may not be available to clients, so cannot be inferred.  We probably should just return void actually

Issue #302, opened 2021-09-17
STORE:  Cannot remove stored account from google


FrozenKiwi:
Login
Navigate to Account Settings => Account Storage
Toggle "Google Store" (on if necessary, first) then off.

What happens:
Alert pops up, says "this is not implemented"

What should happen:'
It should remove the account from Google

PR #303, opened 2021-09-17
STORE: Restore toggle in account settings backs up account correctly

FrozenKiwi:
Triggers this issue:
https://github.com/thecointech/thecoin/issues/302


libs/account/src/state.ts
```diff
@@ -26,7 +26,7 @@ export type AccountState = {
 
 
  // IDX vars
- idx?: IDX;
+ idx: IDX|null;
```
FrozenKiwi, line 29: Try to use `null` for things that are there, but null, and `undefined` for things that are not found

libs/site-base/src/containers/AddAccount/Storage/GDrive/Restore.tsx
```diff
@@ -1,13 +1,21 @@
...
 
 const text = defineMessage({ defaultMessage: "Load from Google Drive", description: "Store Account: Option to store on GDrive" });
 
 export const GDriveRestore = () => {
   const [token, setToken] = useState(undefined as MaybeString);
+  const [loading, doTransmit] = useGoogle()
   return token
     ? <Redirect to={`/addAccount/restore/list?token=${token}`} />
-    : <GDriveBase text={text} onAuth={setToken} />
+    : <ProviderChoice
```
FrozenKiwi, line 15: Separate UI and code

PR #310, opened 2021-09-22
Fix infinte loop on gen account

FrozenKiwi:
When creating an account, it was possible to overflow the stack if submitting before the validation of any of the inputs was complete

### DIFF
--- a/libs/shared/src/components/UX/Input/index.tsx
+++ b/libs/shared/src/components/UX/Input/index.tsx
@@ -46,7 +46,7 @@ export const UxInput = (props: BaseProps) => {
     else {
       localChange(value);
     }
-  }, [forceValidate, onValidate])
+  }, [forceValidate])
 
   // Reset to default value either if requested, or if defaultValue changes
   useEffect(() => {
--- a/libs/shared/src/containers/Account/reducer.ts
+++ b/libs/shared/src/containers/Account/reducer.ts
@@ -47,7 +47,6 @@ function AccountReducer(address: string, initialState: AccountState) {
 
     ///////////////////////////////////////////////////////////////////////////////////
     // Save/load private details
-
     *getIDX() {
       let idx = this.state.idx;
       if (!idx) {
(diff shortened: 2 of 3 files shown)


libs/shared/src/components/UX/Input/index.tsx
```diff
@@ -46,7 +46,7 @@ export const UxInput = (props: BaseProps) => {
     else {
       localChange(value);
     }
-  }, [forceValidate, onValidate])
```
FrozenKiwi, line 49: because onValidate could be a local fn (in some forms) it could change every re-render

FrozenKiwi: if forceValidate was `true`, then the re-render would trigger re-validation, which would then cause a re-render etc.

libs/shared/src/containers/Account/reducer.ts
```diff
@@ -165,26 +164,19 @@ function AccountReducer(address: string, initialState: AccountState) {
           undefined :
           (per: number) => {
             const percent = Math.floor(per * 100);
-            if (!callback(percent)) {
-              throw ("Operation cancelled");
-            }
+            return !callback(percent);
           }
```
FrozenKiwi, line 168: Throwing here somehow escapes the catch below, however we could just return the cb value and let ethers throw for us

libs/shared/src/containers/Login/index.tsx
```diff
@@ -109,6 +109,9 @@ export const Login = (props: Props) => {
       setPercentComplete(-1);
       if (!__cancel) {
         setValidateFn(() => badPwdValid(password!));
+        // Force validation update by toggling forceValidate
+        setForceValidate(v => !v);
+        setForceValidate(v => !v);
```
FrozenKiwi, line 114: Because onValidate is no longer a trigger for re-validation, we must manually force the validation if we change the validate fn.

Technically, this is only necessary if forceValidate is true, (it is always true here)

MarieTheCoin (approved):

PR #312, opened 2021-09-27
Fix/restart deposits

FrozenKiwi:
Not technically a fix, as it wasn't implemented.

Same mechanism for e-Transfers applied to Deposits

### DIFF
--- a/apps/admin/src/containers/BrokerCAD/Clients/ManualOverride.tsx
+++ b/apps/admin/src/containers/BrokerCAD/Clients/ManualOverride.tsx
@@ -1,8 +1,7 @@
 import React, { useState } from 'react';
-import { AnyAction } from 'redux';
-import { ActionType, storeTransition } from '@thecointech/broker-db';
+import { ActionDataTypes, ActionType, AnyAction, AnyActionData, storeTransition } from '@thecointech/broker-db';
 import { graph } from '@thecointech/tx-etransfer';
-import { etransfer } from '@thecointech/tx-deposit';
+import { etransfer, manual } from '@thecointech/tx-deposit';
 import { Button, Select } from 'semantic-ui-react';
 import { log } from '@thecointech/logging';
 import { manualOverrideTransition } from '@thecointech/tx-statemachine/transitions';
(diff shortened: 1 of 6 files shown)


libs/rbcapi/src/action.ts
```diff
@@ -59,7 +60,9 @@ export class ApiAction {
   step: number = 0;
 
   private constructor(identifier: string) {
-    this.outCache = `/temp/TheCoin/logs/rbcapi/Screenshots/${identifier}`;
+    const base = process.env.TC_LOG_FOLDER ?? tmpdir()
```
FrozenKiwi, line 63: hard-coded path broke linux (and all good coding standards)

apps/admin/src/containers/BrokerCAD/Clients/ManualOverride.tsx
```diff
@@ -2,6 +2,7 @@ import React, { useState } from 'react';
 import { AnyAction } from 'redux';
 import { ActionType, storeTransition } from '@thecointech/broker-db';
 import { graph } from '@thecointech/tx-etransfer';
+import { etransfer } from '@thecointech/tx-deposit';
```
FrozenKiwi, line 5: possibly need better names than 'graph' for the exported transition graphs for our actions

apps/admin/src/containers/BrokerCAD/Clients/ManualOverride.tsx
```diff
@@ -22,10 +23,10 @@ export const ManualOverride = (props: AnyAction) => {
   )
 }
 
-function buildOptions(_type: ActionType) {
+function buildOptions(type: ActionType) {
   // What states can we go to?
-  // Assume eTransfer for now
-  const states = Object.keys(graph);
+  // We may need to consider a cleaner view than the raw to-any-state
+  const states = type == "Buy" ? Object.keys(etransfer) : Object.keys(graph);
```
FrozenKiwi, line 29: does not consider manual purchases, as they will probably never fail (can implement it if it ever proves necessary)

MarieTheCoin: "will never fail" makes me really anxious every times I heard that one it failed at some point

apps/admin/src/containers/BrokerCAD/Clients/ManualOverride.tsx
MarieTheCoin: Is it linked to the "visual" graph?

MarieTheCoin (approved):

apps/admin/src/containers/BrokerCAD/Clients/ManualOverride.tsx
FrozenKiwi: you're right.  Implementing that now

PR #336, opened 2021-11-15
Fix/google store no tab

FrozenKiwi:
Backing up an account would fail if `window.open` is not present or did not work for whatever reason.

Discovered by user when accessing an account from within android messenger


FrozenKiwi:
MERGED now so we can get a publish up asap, but still needs review

PR #337, opened 2021-11-18
Fix/fetch history limit

FrozenKiwi:
Tiny job, massive fix.

After moving our smart contract to polygon, the existing getLog we were using would not search spans of more than 10K blocks.  Unfortunately, because Polygon moves so fast, that range was exhausted in a few days.  

Polygon is built on etherscan though so, which has special-purpose methods for scanning blocks.  Solution, switch provider from Infura to use Etherscan and connect directly.

However, the built-in Etherscan provider didn't work with Mumbai.  Luckily, 3rd party plugin provided that ability.  Unfortunately, it required a slightly newer version of Ethers.  No worries, we upgrade our version of ethers.  Unfortunately, this then conflicted with older versions of ethers used by IDX.  I went to update IDX, and found that all my existing tools had been deprecated, and the whole system switched to a new set in the last few months.  Tried to revert, unfortunately, the mix of new & old ethers was now breaking the devLive configuration.  Threw up hands in despair, upgraded to the new version of IDX (is a lot more dev-friendly, would rather waste cycles going forward than backwards.  After everything was working, tested the system vs devlive.  Unfortunately, the new versions of ethers changed the getLogs we were using to communicate with devlive.  The fix to this mismatch was included in the alpha version of ganache 7.

At this point, I threw up my hands.  Created my own provider, wrote bypass code to use the old version of getLogs in devlive, and then re-wrote the transaction fetch code to run using etherscans fancy new apis.  Then back-ported the new api's to work using the old logs to get devLIve going.

Then, cause it was so much fun, I backported the whole system to dev, so can run the website (with transactions) without devLive

God I hate crypto...


environments/devlive.public.env
```diff
@@ -41,8 +41,8 @@ URL_SERVICE_NFT=http://localhost:7003/api/v1
 # The following variables are obviously environment-specific, so probably shouldn't be in GIT
 TC_LOG_FOLDER=/temp/TheCoin/debug/logs
 
-# When we deploy, attempt to clone production state onto this contract
-DEPLOY_CONTRACT_INIT=clone
+# When we deploy, seed some basic tx logic
```
FrozenKiwi, line 44: cloning prod tages ages, was only useful to test the cloning feature

libs/contract-core/contracts/TheCoin.sol
```diff
@@ -65,7 +65,7 @@ contract TheCoin is ERC20Upgradeable, AccessControlUpgradeable {
     // We record the exact timestamp a transaction was initiated
     // to ensure our tracking is precise (for tax etc).
     // NOTE: timestamp here is in millis (not seconds)
-    event ExactTransfer(address from, address to, uint amount, uint timestamp);
+    event ExactTransfer(address indexed from, address indexed to, uint amount, uint timestamp);
```
FrozenKiwi, line 69: The truffle develop console seemed to be unable to connect to our contract if it was nested.

That wasn't really noticeable until I tried using hte dev console to verify what had happened to all the logs

libs/contract-core/package.json
```diff
@@ -32,9 +31,12 @@
     "clean:contracts": "rimraf ./src/contracts",
     "clean:types": "rimraf ./src/types && rimraf ./migrations/types",
     "test:contract": "truffle test",
-    "prod:test": "cross-env CONFIG_NAME=prodtest yarn _deploy",
-    "dev:live": "yarn wait-port 9545 && cross-env CONFIG_NAME=devlive yarn _deploy",
-    "_deploy": "yarn contract-migrate && yarn contract-verify",
+    "prod:test": "cross-env CONFIG_NAME=prodtest yarn deploy",
+    "dev:live": "yarn wait-port 9545 && cross-env CONFIG_NAME=devlive yarn deploy:polygon",
+    "deploy": "runs-s deploy:*",
+    "deploy:polygon": "truffle migrate --network polygon",
```
FrozenKiwi, line 37: I think I got a bit carried away with centralizing scripts here.  Was more efficient to leave deployement in this file

libs/ethers-provider/src/getDefaultApiKey.ts
```diff
@@ -0,0 +1,36 @@
...
+    case 'arbitrum-testnet':
+    case 'homestead':
+    case 'ethereum':
+    case 'mainnet':
+    case 'kovan':
+    case 'ropsten':
+    case 'goerli':
+    case 'rinkeby':
+    case 'optimism-kovan':
+      return '9D13ZE7XSBTJ94N9BNJ2MA33VMAY2YPIRB'
```
FrozenKiwi, line 19: These are not-my-keys (form the package I sourced this from).

I don't use them, I might remove them (done)

FrozenKiwi:
@MarieTheCoin If you feel like reviewing this, I would definitely value your input

I'm going to merge it in now though because I need to deploy the fixes in here.  I'll re-review it myself tomorrow if you don't see this/get around to it.

PR #344, opened 2021-11-30
Prettier setup

MarieLoupG:
If you are interested in having a uniform formatting for everybody: following this should allow you to format on save your files. It's going to pollute the PRs at first but in the long run should be way better.
 - yarn install
 - Install eslint in VSCode
 - Add in VSCode settings.json
`
    "editor.formatOnSave": false,
    "editor.codeActionsOnSave": {
        "source.fixAll.eslint": true
    }
`


FrozenKiwi:
Hey Marie, I've remembered why I removed prettier in the first place - it's very opinionated about line length.  I try to keep one piece of logic per line, eg:
```ts
thing != explode  // This means one thing
  ? crowd = cheer  // This means something else
  : hospital(crowd); // unrelated to above statement
```
I use this kind of thing quite a lot in code to keep my variables `const` as much as possible, and I struggle with reading when it gets collapsed onto a single line.

I get why prettiers so opinionated, but it's kinda a deal-breaker for me.

I think we can get most of what we want (but without forcing the line length) by using eslint though, no?

PR #345, opened 2021-12-07
Fix/final contract transition

FrozenKiwi:
It's the final step: transitioning from our current contract on the test network onto the polygon network.

This involved adding support for hardware wallets, re-writing our deployment to avoid root-level await.  We also centralized our use of providers, and refactored/re-wrote a large part of our contract.

It is hoped that this is the last large-scale refactoring for a very long time


apps/broker-service/src/status/index.ts
```diff
@@ -16,16 +14,9 @@ export async function current() : Promise<Status> {
...
+    // In dev, we can simply load the BrokerCAD to get it's address
+    const brokerCad = await getSigner('BrokerCAD');
+    return brokerCad.getAddress();
   }
-  // Choose the correct account for test/prod
-  else if (process.env.SETTINGS === 'testing') {
-    return prodtest.BrokerCAD;
-  }
-  return production.BrokerCAD;
+  return process.env.WALLET_BrokerCAD_ADDRESS ?? "BROKER_ADDRESS_MISSING";
```
FrozenKiwi, line 21: Our environment variables can be relied on, and we don't want them in 2 places

libs/contract-core/contracts/TheCoin.sol
```diff
@@ -22,281 +17,52 @@ import "@openzeppelin/contracts-upgradeable/token/ERC20/ERC20Upgradeable.sol";
...
-        require(msg.sender == role_Police, "Invalid sender");
-        _;
-    }
-
-    modifier timestampIncreases(address from, uint256 timestamp)
-    {
-        require(timestamp > lastTxTimestamp[from], "Provided timestamp is lower than recorded timestamp");
-        _;
-    }
+contract TheCoin is Pluggable {
```
FrozenKiwi, line 20: I broke down this file into a series of smaller contracts to isolate each piece of functionality into a more testable unit.

Also, re-wrote most of it to support plugins, use roles, and dropped tap-cap-manager

libs/contract-core/seed/devlive/index.ts
```diff
@@ -0,0 +1,56 @@
+import { DateTime } from 'luxon';
+import { log } from '@thecointech/logging';
+import { ConnectContract, TheCoin, COIN_EXP } from '@thecointech/contract-core';
```
FrozenKiwi, line 3: Moved seeding out of migrations, because it uses a different HW wallet to the original, and we can't do both at the same time

PR #351, opened 2021-12-16
Landing/grid layout

FrozenKiwi:
Converts landing to grid.  Mostly updating navigation/footer.


apps/site-app/lang/en.json
```diff
@@ -435,6 +435,62 @@
...
+  "shared.widgets.balanceandprofit.balance": {
+    "defaultMessage": "Balance",
+    "description": "shared.widgets.balanceandprofit.balance: Title for widget Balance and profit"
+  },
+  "shared.widgets.balanceandprofit.cad": {
+    "defaultMessage": "$CAD",
+    "description": "shared.widgets.balanceandprofit.cad: Title for widget Balance and profit"
+  },
+  "shared.widgets.balanceandprofit.profit": {
+    "defaultMessage": "Profit",
```
FrozenKiwi, line 447: We should probably remove this file from git, and figure out a better way to track translations

MarieTheCoin: I am not sure what the other options are. The only other options I know of are just adding interfaces for translators.

libs/site-semantic-theme/src/semantic/thecoin/globals/site.variables
```diff
@@ -28,6 +28,20 @@
 @pageBackground: @white;
 @maxContentWidth: 1440px;
 
+// Scalable settings run from small to big depending on screen size
```
FrozenKiwi, line 31: scalable settings now available globally (and as a fn, instead of a property)

apps/site-app/lang/en.json
FrozenKiwi: Neither - but there -must- be something?  Like, isn't this a problem that everyone everywhere has?  I did look into some BYO translation services, and there was one good one but required switching i18n library (just... no...)

MarieTheCoin: You are probably going to be mad at me for saying this again but: it was a bit better with the fixed ids.

apps/site-landing/src/containers/App/index.tsx
```diff
@@ -20,28 +18,29 @@ import { MediaContextProvider, mediaStyles } from '@thecointech/shared/component
...
 // - LESS is slower, but offers on-save hot-reload
 // - CSS is faster, but requires manual recompile
-import '../../semantic/semantic.css';
-//import '@thecointech/site-semantic-theme/semantic.less';
+//import '../../semantic/semantic.css';
+import '@thecointech/site-semantic-theme/semantic.less';
 import styles from './styles.module.less';
 
 export const App = () => {
   Prismic.useStore();
```
MarieTheCoin, line 28: Maybe this could be put in a context

apps/site-landing/src/containers/Subscribe/styles.module.less
```diff
@@ -1,46 +1,40 @@
...
+  padding-top   : 80px;
+  display       : flex;
+  flex-direction: column;
+  justify-items : center;
+  color         : white;
 
-  .subContainer {
-    margin: auto;
+  .message {
+    margin    : auto;
```
MarieTheCoin, line 11: Not a big fan of those. It was a while back but some browser refused to interpret those.

apps/site-landing/src/semantic/elements/button.variables
```diff
@@ -0,0 +1,12 @@
+
+@secondaryColor: @theCoinPaletteGreen3;
+// This colors should be added to the palette, however I
+// am finding the palette pretty crowded already (ie, not useful).
+// Will leave here until we can figure out a more efficient way
+// to label our colors
+@secondaryColorHover: #42CBA3;
+@secondaryColorDown: #1CA47F;
```
MarieTheCoin, line 8: Agreed but I am not a big fan of having those codes defined again elsewhere

MarieTheCoin (approved):

apps/site-landing/src/containers/MainNavigation/styles.module.less
```diff
@@ -1,63 +1,62 @@
...
-  @{property}: clamp(unit(@small, @unit),
-      calc(unit(@small, @unit) + (@big - @small) * ((100vw - @tabletBreakpoint) / (@desktopWidth - @tabletWidth))),
-      unit(@big, @unit),
-    );
-}
-
-.scalable(@property, @small, @big) {
-  .scalable-unit(@property, @small, @big, px)
+.background {
+  background-color : white;
```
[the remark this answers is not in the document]
FrozenKiwi: fixed

apps/site-landing/src/containers/App/index.tsx
FrozenKiwi: I'm not sure what you mean - like a react context?

apps/site-landing/src/containers/App/tests/__snapshots__/index.test.tsx.snap
```diff
@@ -1,10 +1,9 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
```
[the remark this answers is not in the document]
FrozenKiwi: It's more to test that the page renders than any useful testing of what is rendered.  I think it was disabled for most of our dev life because we were just changing so often

apps/site-app/lang/en.json
FrozenKiwi: :-). There are pro's and con's of each approach.  Mostly I just didn't want to have to update the IDs to be correct after moving the component and didn't want to leave incorrect ids in there.

Translation is a problem I'll have another crack at soon.  We have to get the translations from shared/base merged into the final version at the very least.

apps/site-landing/src/containers/App/index.tsx
MarieTheCoin: Yes. Putting the stores in a context and putting that context parent of everything is really a time saver.

FrozenKiwi: I think that's kind of what we already do - but we do it dynamically so nobody has to -know- ahead of time what's in the store.

We do that so we can split the store over shared/base/site(s) but each library can act as if it's in a sandbox.  It also means that we don't start a store if you don't use it.  This particular store probably should be moved to the "Help" page so if it isn't used it's not even started.  It's not a big deal for most things, but the account saga's are pretty heavy so it's mostly so we don't start them if we don't need them (ie, if you have a bunch of accounts - we only inject the store for the actual account we are using).

At least, that's what we are doing, which AFAIK (caveat - not much) is similar to what you are suggesting?

apps/site-landing/src/containers/Subscribe/styles.module.less
FrozenKiwi: neither.  Also, the auto formatter does a really bad job of LESS mixin formatting.  So will probably ditch it as soon as I have time to look for alternatives

MarieTheCoin: Did you take a look at the PR I opened? https://github.com/thecointech/thecoin/pull/344

apps/site-landing/src/containers/App/index.tsx
FrozenKiwi: (if there is a simple example of what you mean though, I'd love to check it out?  Might be a fancy new technique, and never know it could come in handy)

apps/site-landing/src/containers/Subscribe/styles.module.less
FrozenKiwi: yes, that's where I plan to start :-).  I figured I'll get to a decent checkpoint, then re-format all the files in the repo so we don't get so much noise in PR's

FrozenKiwi:
SNYK failing, but build/tests have passed so merging anyway

PR #352, opened 2021-12-20
Landing/landscape

FrozenKiwi:
Move Landscape into Grid layout.

Still not animated, but hopefully can get to that soon.

I couldn't figure out how to get the images to line up nicely, so I re-exported them all at max width (1440px).  That means we can anchor them easily & they don't get out of sync with each other.

https://user-images.githubusercontent.com/4876160/146997929-d52a5d46-27ae-465f-8309-5538d8a38b45.mp4


apps/site-landing/src/containers/HomePage/landscape/Background.tsx
```diff
@@ -0,0 +1,7 @@
+import React from 'react';
+import styles from './background.module.less';
+
+export const Background = () => (
+  <div className={styles.background}>
```
FrozenKiwi, line 5: now I'm looking at this again, I don't know why I put background in it's own div.

I don't want to change it now (I've spent too long on this already) but maybe future-me can try merging the background styles onto the content div

apps/site-landing/src/containers/HomePage/landscape/background.module.less
```diff
@@ -0,0 +1,44 @@
...
+    normal;
+
+  background-repeat: no-repeat;
+
+  .smallScreen({
+    background-image: url('./images/mobile.svg');
+    height: 1165px;
+    background-position-x: 48%;
+    background-repeat: no-repeat;
+  })
```
FrozenKiwi, line 43: a lot of work for something that is still static... *sigh*

PR #355, opened 2021-12-22
Landing/cta bottom

FrozenKiwi:


apps/site-landing/src/containers/App/styles.module.less
```diff
@@ -18,7 +18,7 @@
     grid-template-areas:
       "lefth header righth"
       "main  main   main"
-      "leftf footer rightf"
+      "lgap footer rgap"
```
FrozenKiwi, line 21: I've suddenly realized I setup the mobile landing without ensuring my base grid was correct (it's not)

FrozenKiwi: I don't care (for now), as it's fairly low on the list of priorities

apps/site-landing/src/containers/CreateAccountBanner/index.tsx
```diff
@@ -12,9 +9,9 @@ export enum TypeCreateAccountBanner {
     Plants
 }
 
-const images = {
-  [TypeCreateAccountBanner.People]: illustrationPeople,
-  [TypeCreateAccountBanner.Plants]: illustrationPlant,
+const bgStyles = {
+  [TypeCreateAccountBanner.People]: styles.withPeople,
+  [TypeCreateAccountBanner.Plants]: styles.withPlants,
```
FrozenKiwi, line 14: I feel I should make background images part of the style rather than a html `img`.  Better for accessibility etc.  Is that true?

apps/site-landing/src/containers/CreateAccountBanner/styles.module.less
```diff
@@ -1,3 +1,29 @@
-#createAccountBanner.content{
+@import "~@thecointech/site-semantic-theme/variables.less";
+
+.content {
```
FrozenKiwi, line 3: Removed ID because we have multiple CreateAccount banners (even tho they aren't the same element, they could become so in future)

apps/site-landing/src/containers/HomePage/underwater/styles.module.less
```diff
@@ -25,7 +26,7 @@
 
   .body {
     margin-top    : -8px;
-    padding-top   : .scalable(20, 150)[@result];
+    padding-top   : .scalable-clamped-bp(50, 124, @mobileWidth, @midWidth, px)[@result];
```
FrozenKiwi, line 29: I do worry that putting this much math into my CSS will end up killing page performance.

libs/site-semantic-theme/src/semantic/thecoin/globals/site.variables
```diff
@@ -40,21 +40,18 @@
 @maxContentWidth: 1440px;
 
 // Scalable settings run from small to big depending on screen size
-.scalable-unit-bp(@small, @big, @bp, @unit) {
-  @result: calc(unit(@small, @unit) + (@big - @small) * ((100vw - unit(@bp, @unit)) / (@desktopWidth - @bp)))
+.scalable-unit-bp(@small, @big, @smallbp, @bigbp, @unit) {
+  @result: calc(unit(@small, @unit) + (@big - @small) * ((100vw - unit(@smallbp, @unit)) / (@bigbp - @smallbp)))
```
FrozenKiwi, line 44: allow setting breakpoints other than mid/max

PR #359, opened 2021-12-23
Admin/incomplete page

FrozenKiwi:


apps/admin/src/init/accounts.ts
```diff
@@ -25,7 +25,7 @@ export async function initialAccounts() {
   return {
     active: null,
     map: {
-      ...(await buildMapEntry("TheCoin")),
+      // ...(await buildMapEntry("TheCoin")),
```
FrozenKiwi, line 28: Requires fixing.  At some future point when we figure out a solution...

PR #368, opened 2022-01-13
Widget/slider

FrozenKiwi:


apps/site-app/package.json
```diff
@@ -71,10 +71,8 @@
     "react-helmet": "6.0.0-beta",
     "react-measure": "^2.3.0",
     "react-semantic-ui-range": "^0.7.0",
-    "react-transition-group": "^4.3.0",
     "reselect": "4.0.0",
-    "sanitize.css": "10.0.0",
-    "ts-clipboard": "^1.0.17"
```
FrozenKiwi, line 77: noticed a lot of unused packages here. They gone now...

apps/site-landing/lang/en.json
```diff
@@ -341,10 +341,6 @@
     "defaultMessage": "Success: check your emails",
     "description": "site.subscribe.success: Message we give a user when the subscription is a success"
   },
-  "tk3UgV": {
```
FrozenKiwi, line 344: text should not be transalted

apps/site-landing/src/components/RangeFieldAndScale/index.tsx
```diff
@@ -0,0 +1,108 @@
...
+          <FormattedNumber
+            value={value}
+            currency={props.currency}
+            currencyDisplay={"narrowSymbol"}
+            style={props.scaleType}
+            minimumFractionDigits={0}
+            maximumFractionDigits={0}
+           />
+        </div>
+        <Range
```
FrozenKiwi, line 47: I couldn't get the in-build one to color the track non-uniformly, so installed a new version to do it.

apps/site-landing/src/containers/Compare/FormCompare/index.tsx
```diff
@@ -26,55 +23,31 @@ const translations = defineMessages({
...
+    <div id={styles.variablesContainer}>
+
+      <RangeFieldAndScale
+        label={translations.startingValue}
+        scaleType="currency"
+        currency="CAD"
+        maximum={1000}
+        step={1}
+        onChange={onChange}
+      />
```
FrozenKiwi, line 39: This file isn't the topic of review.  Updated to compile, doesn't matter if it 'works'

libs/site-semantic-theme/src/semantic/thecoin/elements/input.overrides
```diff
@@ -8,12 +8,12 @@
 // definition could override this one
 .ui.input {
 
-  & > input {
+  & > input[type="text"] {
     font-weight: @bold;
 
     &:hover {
        border-color: @hoverBorderColor;
-       box-shadow: @hoverBoxShadow !important;
```
FrozenKiwi, line 16: important overrode the shadows on our range input, so had to apply it a different way

apps/site-landing/lang/en.json
MarieTheCoin: Transalted: for a second I thought I was learning a new word today

apps/site-landing/src/components/RangeFieldAndScale/Slider.stories.tsx
```diff
@@ -0,0 +1,34 @@
...
+      default: 'graphColor',
+      values: [
+        { name: 'graphColor', value: '#138175' },
+      ],
+    },
+  },
+  argTypes: { onChange: { action: 'onChange' } },
+  args: {
+    label: "Label",
+    currency: "CAD" as "CAD",
```
MarieTheCoin, line 20: That's a surprising line. What is it doing?

MarieTheCoin (approved):

apps/site-landing/src/components/RangeFieldAndScale/Slider.stories.tsx
FrozenKiwi: ha!  Yeah, its kinda WTF :-).

Our slider displays a currency. One day we might have multiple currencies, so it accepts an argument (which kind of currency).  But we want to type-check the type of currency, so the only one accepted is type "CAD".  But when we put the argument here of "CAD", TS thinks it's of type string unless I explicitly override it with type "CAD".

If I put "CAD" directly as argument to component, I wouldn't need it, it's just cause I let it come in as an argument.

All technically correcter, but prolly more work than it's worth.

PR #369, opened 2022-01-14
tweaking e-Transfer submission to ensure consistency between front/ba…

FrozenKiwi:
…ck ends

### DIFF
--- a/apps/site-app/src/containers/Accounts/Redeem/RedeemWidget/index.tsx
+++ b/apps/site-app/src/containers/Accounts/Redeem/RedeemWidget/index.tsx
@@ -47,6 +47,10 @@ const translations = defineMessages({
   entryIsRequired : {
       defaultMessage: 'This field is required',
       description: 'Error on attempted submission without completing field'},
+  entryTooShort: {
+    defaultMessage: 'This entry needs to be at least {count} letters long',
+    description: 'Error on attempted submission without completing field'},
+
   noSpecialCaractDesc : {
       defaultMessage: 'No numbers or special characters',
       description: 'app.accounts.redeem.form.noSpecialCaractDesc: Label for the form the make a payment page / etransfert tab'},
@@ -90,9 +94,17 @@ type VisualProps={
 
 };
 
-const validateChars = (re: () => RegExp) => (value: string) => {
-  if (!value?.length) {
-    return translations.entryIsRequired
+const validateChars = (re: () => RegExp, minimumLength: number) => (value: string) => {
+  if (minimumLength > 0 && !value?.length) {
+    return translations.entryIsRequired;
+  }
+  if (value.length < minimumLength) {
+    return {
+      ...translations.entryTooShort,
+      values: {
+        count: minimumLength
+      }
+    }
   }
   const match = re().exec(value);
   return match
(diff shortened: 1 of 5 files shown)


apps/site-app/src/containers/Accounts/Redeem/RedeemWidget/index.tsx
```diff
@@ -90,9 +94,17 @@ type VisualProps={
...
 };
 
-const validateChars = (re: () => RegExp) => (value: string) => {
-  if (!value?.length) {
-    return translations.entryIsRequired
+const validateChars = (re: () => RegExp, minimumLength: number) => (value: string) => {
+  if (minimumLength > 0 && !value?.length) {
+    return translations.entryIsRequired;
+  }
+  if (value.length < minimumLength) {
```
FrozenKiwi, line 101: Split length into it's own for better user feedback

apps/site-app/src/containers/Accounts/Redeem/RedeemWidget/index.tsx
```diff
@@ -104,8 +116,9 @@ const validateChars = (re: () => RegExp) => (value: string) => {
       }
     : null;
 }
-const validateAnswer = validateChars(invalidAnswer);
-const validateQuestion = validateChars(invalidQuestion);
+const validateAnswer = validateChars(invalidAnswer, 3);
+const validateQuestion = validateChars(invalidQuestion, 3);
+const validateMessage = validateChars(invalidQuestion, 0);
```
FrozenKiwi, line 121: Message does not have a minimum length, this fixes the error flash on submission too

libs/shared/src/components/UX/Email/index.tsx
```diff
@@ -21,10 +22,9 @@ export const UxEmail = (props: Props) => {
...
   const { onValidate, tooltip, placeholder, intlLabel, ...rest } = props;
   const addressValidate: ValidateCB = (value) => {
-    const split = value?.indexOf('@');
-    if (split > 0 && (split + 1) < value?.length)
-      return onValidate?.(value) ?? null;
-    return localFormat;
+    return value?.match(validEmail())
+      ? onValidate?.(value) ?? null
+      : localFormat;
   }
```
FrozenKiwi, line 28: Consistent email validation.

NOTE: this actually accepts email addresses that RBC refuses.  Not keen on duplicating errors into our code because RBC Is too lazy to read the email spec.

libs/tx-etransfer/src/transitions/sendETransfer.ts
```diff
@@ -71,17 +71,3 @@ function decryptInstructions(packet: EncryptedPacket) {
   return decryptTo<ETransferPacket>(privateKey, packet);
 }
 
-//
```
FrozenKiwi, line 74: This only ran on the server.  Moved into common code so we can check prior to submit *just in case*.  It would have caught several issues missed client-side recently

libs/utils-ts/src/VerifiedSale.ts
```diff
@@ -6,6 +6,9 @@ import { BuildVerifiedAction } from "./VerifiedAction";
 
 // We cannot use the following characters in the question/answer
 // Invalid characters: < or >, { or }, [ or ], %, &, #, \ or "
+export const validQuestion  = () => /^[^\<\>\{\}\[\]\%\&\#\\\"]{3,}$/g;
+export const validAnswer    = () => /^[^\<\>\{\}\[\]\%\&\#\\\"\s]{3,}$/g;
+export const validEmail     = () => /^[^@\s]+@[^@\s]+$/
 export const invalidAnswer   = () => /[\<\>\{\}\[\]\%\&\#\\\"\s]/g;
```
FrozenKiwi, line 12: I don't like duplicating the regex's here, but I still want the invalid regex so we can give clearer feedback to the user (validity is pass/fail in the above regexs)

libs/tx-etransfer/src/transitions/sendETransfer.ts
MarieTheCoin: Always better to double check 👍🏻

MarieTheCoin (approved):

PR #371, opened 2022-01-14
Fix/coin deposit

FrozenKiwi:
Coin transfers should affect their internal balance on completion, not on initialization.

This allows us to compensate for lost transactions in periods of very high traffic.  It also makes a lot more sense to affect it then, as that's the point where we *know* the transaction completed.

Tested with basic jest test, and by running deposit & withdrawal in devlive


libs/contract-core/package.json
```diff
@@ -48,13 +48,21 @@
   ],
   "main": "./build/index.js",
   "exports": {
+    "./contract": "./build/cjs/index.js",
```
FrozenKiwi, line 51: very shonky work-around lets us reference production contract in tests.

FrozenKiwi: (cause we can't validate code that connects to ext servers without it)

libs/tx-statemachine/src/transitions/depositCoin.ts
```diff
@@ -35,7 +34,6 @@ const doDepositCoin: TransitionCallback<BSActionTypes> = async (container) => {
   return (tx.hash)
     ? {
         hash: tx.hash,
-        coin: new Decimal(value),
```
FrozenKiwi, line 38: No!  Bad!  We only sent the transaction!  We have no idea if it completed!

libs/tx-statemachine/src/transitions/waitCoin.ts
```diff
@@ -14,15 +16,49 @@ export async function waitCoin(container: AnyActionContainer) {
...
+
+function parseLog(contract: TheCoin, log: TransactionReceipt["logs"][0]) {
+  try {
+    return contract.interface.parseLog(log)
+  }
+  catch {
+    return null;
+  }
+}
+export function updateCoinBalance(container: AnyActionContainer, receipt: TransactionReceipt) {
```
FrozenKiwi, line 35: Instead of just assuming a transaction that was accepted also completed, here we get the actual logs and fetch what was transferred from here

libs/contract-core/package.json
MarieTheCoin: Can this be used for e2e tests?

libs/tx-statemachine/src/transitions/waitCoin.ts
MarieTheCoin: Is it saved in the db?

MarieTheCoin (approved):

libs/contract-core/package.json
FrozenKiwi: Not quite the same: even E2E tests would normally only run under devlive; so no external connections.  Here I just wanted to be able to explicitly test the external code.

However, obvs your also right (and I haven't thought of this) that our E2E tests need to run under devlive, which is a very different environment to test.

Some thinking required...

libs/tx-statemachine/src/transitions/waitCoin.ts
FrozenKiwi: in this case, logs means eth logs, (not the same thing as regular logs).

Every action on the chain emits a log, so it's a secure method ot test if something happened.

FrozenKiwi: in ethereum, you can only directly query the current state.  If you need historical info, you can only find it in the logs

apps/tx-processor/src/etransfer/index.test.mockdb.json
```diff
@@ -2,13 +2,13 @@
...
     {
       "id":  "0xd918c1b0559a68a78a8d398fd00bcb481470d221105228fdea2a4bd798b5f6f2",
-      "ref": "User/0xCA8EEA33826F9ADA044D58CAC4869D0A6B4E90E4/Sell/0xd918c1b0559a68a78a8d398fd00bcb481470d221105228fdea2a4bd798b5f6f2"
+      "ref": "User/0x445758E37F47B44E05E74EE4799F3469DE62A2CB/Sell/0xd918c1b0559a68a78a8d398fd00bcb481470d221105228fdea2a4bd798b5f6f2"
     }
 
   ],
   "User": [
     {
-      "id": "0xCA8EEA33826F9ADA044D58CAC4869D0A6B4E90E4",
```
FrozenKiwi, line 11: normalized sample code on testing address

libs/contract-core/src/index_mocked.ts
```diff
@@ -7,44 +7,83 @@ const genRanHex = (size: number) => [...Array(size)].map(() => Math.floor(Math.r
...
+let nonce = 12;
+let confirmations = 1;
+export class TheCoin implements Pick<Src.TheCoin, 'exactTransfer' | 'balanceOf' | 'certifiedTransfer'>{
 
-  genReceipt = (opts?: any) => Promise.resolve({
-    wait: () => { },
-    hash: `0x${genRanHex(64)}`,
-    ...opts,
-  } as any as ContractTransaction)
+  // prefix is (e) for exactTransfer and (c) for certifiedTransfer, and (0) for other
```
FrozenKiwi, line 14: This is becoming very convoluted, and is mostly unnecessary.

We can instead standardize on a mocked provider and then just use the regular contract

https://github.com/thecointech/thecoin/issues/376

PR #388, opened 2022-01-31
Landing/benefits graph

FrozenKiwi:


apps/site-landing/src/containers/BenefitsSimulator/Graph/index.tsx
```diff
@@ -0,0 +1,79 @@
...
+
+  useEffect(() => {
+    if (!snpData) return;
+    let isCancelled = false;
+    const cancellableProgress = (value: number) => {
+      setProgress(value);
+      return !isCancelled;
+    }
+    const allReturns = calcAllReturns(snpData, params, cancellableProgress);
+    if (allReturns) {
```
FrozenKiwi, line 38: possible UX-improvement.  Instead of blocking for minutes at a time, we could use generators to start displaying the results as quickly as possible

PR #390, opened 2022-02-04
Landing/simulate article

FrozenKiwi:
This implements a simulator for TheCoin.

It's job is to simulate the performance of the coin using historical data to show people how small changes can have such big results


MarieTheCoin (approved):
Don't know if it's normal I wasn't able to make it work.

FrozenKiwi:
> Don't know if it's normal I wasn't able to make it work.

I haven't integrated it into the website yet - this is just the engine, no bodywork.

PR #392, opened 2022-02-04
eliminate duplicate 'delay' implementations

FrozenKiwi:
We had the same utility function in about 6 places, and I needed it again so refactored all existing ones into utilities

### DIFF
--- a/libs/async/src/index.ts
+++ b/libs/async/src/index.ts
@@ -1,4 +1,3 @@
-
 // Yield execution
 export function sleep(ms: number) {
   return new Promise(resolve => setTimeout(resolve, ms));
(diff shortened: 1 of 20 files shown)


FrozenKiwi:
probably doesn't need a review.  If it builds, it's good.

libs/async/src/index.ts
```diff
@@ -0,0 +1,4 @@
+// Yield execution
+export function sleep(ms: number) {
```
MarieTheCoin, line 2: Is it the only thing in /async?

MarieTheCoin (approved):

libs/async/src/index.ts
FrozenKiwi: yeah... I originally put it in `utilities` but then I messed up the dependency graph.

Having a whole library for a 1-line function is the definition of overkill.  I hope at some point I add some more functions to this to justify the effort.  Buf if I'm forced to split out utility functions into their own libraries, it's probably better to group by functionality so I don't have to do this again in the future.

PR #393, opened 2022-02-05
Landing/integrate graph

FrozenKiwi:
I could watch this simulator all day...

https://user-images.githubusercontent.com/4876160/152660129-763a88c2-5fe4-4496-b365-cf6855117109.mp4

### DIFF
--- a/apps/site-landing/src/containers/BenefitsSimulator/Graph/BenefitsGraph.stories.tsx
+++ b/apps/site-landing/src/containers/BenefitsSimulator/Graph/BenefitsGraph.stories.tsx
@@ -2,6 +2,7 @@ import React, { useEffect, useState } from 'react';
 import { Meta, Story } from '@storybook/react';
 import { BenefitsGraph as Component } from '.';
 import { createParams, MarketData } from '../../ReturnProfile/data';
+//@ts-ignore
 import { generateData } from '../../../../internals/historical/simulation';
 
 const meta: Meta = {
--- a/apps/site-landing/src/containers/BenefitsSimulator/Graph/BenefitsGraph.stories.tsx
+++ b/apps/site-landing/src/containers/BenefitsSimulator/Graph/BenefitsGraph.stories.tsx
@@ -11,6 +11,7 @@ const meta: Meta = {
   args: {
     loadingDelay: 2, // simulated delay of loading data.
     growthPercent: 9,
+    yearsToDisplay: 10,
     yearsToSimulate: 60,
     noise: 0.1,
   }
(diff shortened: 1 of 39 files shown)


apps/site-landing/src/containers/BenefitsSimulator/Graph/index.tsx
```diff
@@ -1,71 +1,118 @@
...
-    }
+    <Segment>
+      <Dimmer active={isLoading}>
+        <GraphLoading progress={progress} />
+      </Dimmer>
+      <AreaGraph maxGraphPoints={maxGraphPoints} data={displayData} />
     </Segment>
-  );
+  )
+  // return isLoading
```
MarieTheCoin, line 106: Can this be removed?

apps/site-landing/src/containers/ReturnProfile/data/params.ts
```diff
@@ -35,7 +35,7 @@ type ShockAbsorber = {
 // The data we use to calculate a clients potential returns
 export type SimulationParameters = {
   // How many months should we be calculating this for?
-  maxDuration: DurationObject;
+  //maxDuration: DurationObject;
```
MarieTheCoin, line 38: Can this be removed?

MarieTheCoin (approved):
I created a PR to fix the styles for the form. It's a temp fix just in case and the mobile version needs love.

apps/site-landing/src/containers/BenefitsSimulator/Graph/BenefitsGraph.stories.tsx
```diff
@@ -2,14 +2,17 @@ import React, { useEffect, useState } from 'react';
 import { Meta, Story } from '@storybook/react';
 import { BenefitsGraph as Component } from '.';
 import { createParams, MarketData } from '../../ReturnProfile/data';
+//@ts-ignore
```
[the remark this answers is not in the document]
FrozenKiwi: yes/no.  It shows up as an error in VS Code, but it actually isn't (it's reading a source file outside of the normal directory, but that doesn't apply to stories).  The `//ts-ignore` just makes it so VSCode doesn't show the file as red

apps/site-landing/src/containers/BenefitsSimulator/Graph/index.tsx
FrozenKiwi: yes, done

apps/site-landing/src/containers/ReturnProfile/data/params.ts
FrozenKiwi: yeah, of course.  It seems pretty unlikely we'll go back to the old way

FrozenKiwi:
> I created a PR to fix the styles for the form. It's a temp fix just in case and the mobile version needs love.

You're a legend Marie :-).

If you don't mind, I'll approve and merge that PR immediately?  I'm currently working on adding a -lot- more parameters to the simulator, so it'd be better to have it in sooner rather than later.

PR #394, opened 2022-02-06
Styles for the graph form

MarieTheCoin:
Added padding on the form, added margin on the slide and removed the button (pretty sure it's useless right now unless the goal is to connect it maybe not a bad idea(?))

Capture d’écran, le 2022-02-06 à 17 55 25

### DIFF
--- a/apps/site-landing/src/components/RangeFieldAndScale/index.tsx
+++ b/apps/site-landing/src/components/RangeFieldAndScale/index.tsx
@@ -1,17 +1,17 @@
 import * as React from 'react';
 import { useState } from 'react';
-import styles from './styles.module.less';
 import { FormattedMessage, FormattedNumber, MessageDescriptor } from 'react-intl';
 import type { NumberFormatOptionsStyle } from '@formatjs/ecma402-abstract';
 
-import { Range, getTrackBackground } from "react-range";
-import { LessVars } from "@thecointech/site-semantic-theme/variables";
+import { Range, getTrackBackground } from 'react-range';
+import { LessVars } from '@thecointech/site-semantic-theme/variables';
+import styles from './styles.module.less';
 
 export type Props = {
  label: MessageDescriptor,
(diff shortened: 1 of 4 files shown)


apps/site-landing/src/components/RangeFieldAndScale/index.tsx
```diff
@@ -1,17 +1,17 @@
...
 import { useState } from 'react';
-import styles from './styles.module.less';
 import { FormattedMessage, FormattedNumber, MessageDescriptor } from 'react-intl';
 import type { NumberFormatOptionsStyle } from '@formatjs/ecma402-abstract';
 
-import { Range, getTrackBackground } from "react-range";
-import { LessVars } from "@thecointech/site-semantic-theme/variables";
+import { Range, getTrackBackground } from 'react-range';
+import { LessVars } from '@thecointech/site-semantic-theme/variables';
+import styles from './styles.module.less';
```
FrozenKiwi, line 8: I like that you pay attention to the small things :-).  I always find it weird when `styles` isn't at the bottom of the import list too.  Is it silly?  Yes.  Does it matter?  No.  Do we do it anyway?  of course...

FrozenKiwi (approved):

apps/site-landing/src/components/RangeFieldAndScale/index.tsx
```diff
@@ -20,52 +20,52 @@ export type Props = {
...
-          renderTrack={(props) => (
-            <SliderTrack
-              value={value}
-              max={maximum}
-              min={minimum}
-              {...props}
-            />)}
-          renderThumb={SliderThumb}
+  return (
+    <div className={`${styles.container} x6spaceBefore`}>
```
FrozenKiwi, line 35: In my merge, I've moved this x6spaceBefore back into the owning (`<Basic />`) class.  I need to squeeze a lot more sliders into the advanced pane and this much margin is top of the chopping block :-).

PR #395, opened 2022-02-07
Landing/advanced sim params

FrozenKiwi:
Full simulation for a complete picture.

The usage is obvs still pretty janky, with the update happening too quickly.  The UI is also very ick.  However, it is working, and incredibly fascinating.

For a look at the power of compounding interest, check out the end of the video where I switch from saving $9 a week (which earns an average of ~$500 per year) to spending 30 per week more than earnings (ends up costing $190K in credit card interest over 8 years).

This is also the first (visible) display of the account drop protection.

https://user-images.githubusercontent.com/4876160/152868922-aee7c9bc-676f-4d7c-b7c5-20e682dc8c4e.mp4

### DIFF
--- a/apps/site-landing/src/containers/BenefitsSimulator/Form/Advanced.tsx
+++ b/apps/site-landing/src/containers/BenefitsSimulator/Form/Advanced.tsx
@@ -55,12 +55,12 @@ const translations = defineMessages({
     description: 'Parameters for simulating the balance protection feature'
   },
   cushion: {
-    defaultMessage: 'Protection Level:',
+    defaultMessage: 'Drop Protect:',
     description: 'How much of a market drop would this account be protected from',
   },
   maximumProtected: {
-    defaultMessage: 'Maximum Balance Protected:',
-    description: 'The maximum balance of the account to protect',
+    defaultMessage: 'Balance Protect:',
+    description: 'The maximum balance to protect with shock absorber',
   }
 });
 
--- a/apps/site-landing/src/containers/BenefitsSimulator/Form/Advanced.tsx
+++ b/apps/site-landing/src/containers/BenefitsSimulator/Form/Advanced.tsx
@@ -3,10 +3,11 @@ import { RangeFieldAndScale } from '../../../components/RangeFieldAndScale';
 import { defineMessages, FormattedMessage, MessageDescriptor } from 'react-intl';
 import { Props } from './types';
 import { debounce } from 'lodash';
-import { PeriodicalParams, SimulationParameters } from '../../ReturnProfile/data/params';
+import { PeriodicalParams, ShockAbsorber, SimulationParameters } from '../../ReturnProfile/data/params';
 import styles from './styles.module.less';
 import { Accordion, Icon } from 'semantic-ui-react';
 import { basic } from './Basic';
+import { Decimal } from 'decimal.js-light';
 
 const translations = defineMessages({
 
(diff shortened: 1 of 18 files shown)


apps/site-landing/src/containers/BenefitsSimulator/Form/Advanced.tsx
```diff
@@ -0,0 +1,276 @@
...
+
+type GroupProps<T> = {
+  title: MessageDescriptor;
+  index: number;
+  activeIndexState: UseNumberState;
+  params: T;
+  setParams: (v: T) => void;
+}
+
+const PeriodicalGroup: React.FC<GroupProps<PeriodicalParams>> = ({ title, index, params, setParams, activeIndexState, children }) => {
```
FrozenKiwi, line 162: This file actually contains 3 components.  Perhaps it should be refactored into a sub-directory and 1 file each component?

apps/site-landing/src/containers/ReturnProfile/data/article.test.ts
```diff
@@ -38,12 +38,12 @@ it ('Matches article with ShockAbsorber', () => {
       // We protect up to $500
       maximumProtected: new Decimal(500),
       // for a max drop of 50%
-      absorbed: new Decimal(0.50),
+      cushionDown: new Decimal(0.50),
```
FrozenKiwi, line 41: I couldn't remember what up/down, renamed because both up/down operations have the same intended effect (to smooth up the jumps/valleys)

apps/site-landing/src/containers/ReturnProfile/data/params.ts
```diff
@@ -51,7 +47,7 @@ export type SimulationParameters = {
   // How much money flows in each week
   // any monthly income (eg rent, child benefits)
   // Any yearly income (eg bonus's, etc)
-  income: FinancialParams;
+  income: PeriodicalParams;
```
FrozenKiwi, line 50: I hate naming things...

apps/site-landing/src/containers/ReturnProfile/data/state.ts
```diff
@@ -73,6 +73,10 @@ export const increment = (state: SimulationState, date: DateTime) => ({
 export const toFiat = (coin: Decimal, data: MarketData) => coin.mul(data.P)
 export const toCoin = (coin: Decimal, data: MarketData) => coin.div(data.P)
 
-export const calcFiat = (state: SimulationState, data: MarketData[]) =>
+export const grossFiat = (state: SimulationState, data: MarketData[]) =>
```
FrozenKiwi, line 76: gross in finance means raw number without costs (in the sim, the credit being carried is the cost)

apps/site-landing/src/containers/BenefitsSimulator/Form/Advanced.tsx
MarieTheCoin: Would be nice and I am absolutely not saying that because I spend several days fighting with a google api and that I am angry at every single developer on earth.

apps/site-landing/src/containers/BenefitsSimulator/Form/Advanced.tsx
```diff
@@ -0,0 +1,276 @@
...
+      <Accordion.Content active={active} >
+        <RangeFieldAndScale
+          label={translations.weekly}
+          scaleType="currency"
+          currency="CAD"
+          maximum={1000}
+          step={1}
+          initial={params.weekly}
+          onChange={setNamedValue("weekly")}
+        />
```
MarieTheCoin, line 195: Those three range fields could just be one and three radio buttons.

apps/site-landing/src/containers/ReturnProfile/data/params.ts
MarieTheCoin: Is that why you rename things so often? 😂

MarieTheCoin (approved):
I don't understand what any of those field do or mean but it looks cool.

apps/site-landing/src/containers/BenefitsSimulator/Form/Advanced.tsx
FrozenKiwi: I think all 3 should be there.  They mean different things.  Fer example - I (theoretically) get paid weekly, child benefits arrive monthly, and once per year I get a bonus.  Then I buy groceries weekly, I pay rent monthly, and I go on vacation once per year.

It might sound pedantic, but I want this to be as accurate as possible.  Also, these things have surprisingly large effects over time.  Getting paid weekly rather than monthly means maybe $10K more in savings at retirement.

Play with the simulator a bit and you'll see what I mean.  (Also, that it's a huge pain to play with).

FrozenKiwi:
> I don't understand what any of those field do or mean but it looks cool.

Help docs are definitely required

PR #402, opened 2022-02-21
updated assets to latest CSS version

FrozenKiwi:
Fixes all the missing assets in last prodtest deployment

### DIFF
--- a/apps/site-landing/internals/server.mjs
+++ b/apps/site-landing/internals/server.mjs
@@ -4,9 +4,10 @@ import { join, dirname } from "path";
 import { readFileSync } from "fs";
 import { fileURLToPath } from "url";
 
-const __filename = fileURLToPath(import.meta.url);
-const __dirname = dirname(__filename);
-const datapath = join(__dirname, '..', 'src', 'containers', 'BenefitsSimulator', 'simulator');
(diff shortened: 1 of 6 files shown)


apps/site-landing/internals/server.mjs
```diff
@@ -4,9 +4,10 @@ import { join, dirname } from "path";
 import { readFileSync } from "fs";
 import { fileURLToPath } from "url";
 
-const __filename = fileURLToPath(import.meta.url);
-const __dirname = dirname(__filename);
-const datapath = join(__dirname, '..', 'src', 'containers', 'BenefitsSimulator', 'simulator');
+const url = import.meta.url;
```
FrozenKiwi, line 7: `unexpected reserved word` when passing import.meta.url directly to fileUrlToPath

libs/site-base/internal/server/middlewares/addProdMiddlewares.mjs
```diff
@@ -1,5 +1,5 @@
 import { resolve } from 'path';
-import { static } from 'express';
+import { static as exstatic} from 'express';
```
FrozenKiwi, line 2: static is reserved keyword too

PR #404, opened 2022-02-22
adding symbol to types

FrozenKiwi:
Add 'types' back to numeric input.

image


libs/shared/src/components/UX/Numeric/index.tsx
```diff
@@ -16,34 +17,55 @@ const errTooHigh = defineMessage({
...
       onValue={toNumValue}
-      defaultValue={defaultValue?.toString()}
+      defaultValue={convertIn(defaultValue, mult)}
+      type='number'
       {...rest}
     />
   );
 }
+
+const convertIn = (value: number|undefined, mult: number) =>
```
FrozenKiwi, line 58: percentages display as "50%" but internally we want normal multiplier (0.5)

libs/shared/src/components/UX/Numeric/styles.module.less
```diff
@@ -0,0 +1,43 @@
+@import "~@thecointech/site-semantic-theme/variables.less";
+
+.numeric {
+  :global(.ui.input)>input {
+    text-align: right;
```
FrozenKiwi, line 5: I'm not 100% convinced it is a good idea to enforce text-align here.  It looks much better, but it can be awkward to select

libs/shared/src/components/UX/Numeric/styles.module.less
```diff
@@ -0,0 +1,43 @@
...
+    }
+  }
+  &.percent span::after {
+    content: '%'
+  }
+  &.currency span::after {
+    content: '$'
+  }
+  &.years span::after {
+    content: 'yrs';
```
FrozenKiwi, line 41: note, un-translatable content here.  Not sure how to handle that?

FrozenKiwi:
merged it in, cause some features needed on test.  However, still would welcome a review :-)

PR #405, opened 2022-02-23
Landing/embed newsletter

FrozenKiwi:
Switched to embedded MailJet newsletter registration.

This means lots of code is now obsoleted & removed - sorry :-(. 
But it also means lots of code deleted - check out that +/- ratio!!! woo :-)

https://user-images.githubusercontent.com/4876160/155415454-b624d9ef-fe26-4955-8d95-278d2284035f.mp4


apps/site-landing/internals/mailjet/newsletter-form.html
```diff
@@ -0,0 +1,105 @@
+<style>
+  #container {
+    display: flex;
+    justify-content: center;
+    line-height: normal;
+    text-align: left;
+    background-color: transparent;
+    box-shadow: none;
```
FrozenKiwi: I used this file to write the HTML for the embedded widget.  It's not really necessary, and violates DRY, so removing it

apps/site-landing/src/containers/App/Routes.tsx
```diff
@@ -16,7 +15,6 @@ import { Privacy } from 'containers/Privacy';
 export const Routes = () =>
   <Switch>
     <Route path="/" exact component={HomePage} />
-    <Route path="/newsletter/confirm" exact component={Confirm} />
```
FrozenKiwi, line 19: This is a questionable decision - It would be nice to host the confirmation page in our website (rather than on mailjet).

Unfortunately, it's probably just so far down the list of priorities that by the time I get to it this code branch would be rotted and more hassle than it's worth

libs/site-semantic-theme/src/semantic/thecoin/elements/button.overrides
```diff
@@ -16,5 +16,5 @@
     }
 }
 .ui.teal.buttons .button, .ui.teal.button{
-        color: #2B5250 !important;
+        color: @theCoinPaletteGreen5 !important;
```
FrozenKiwi, line 19: random fix - stumbled on hard-coded color and replaced it.

PR #407, opened 2022-02-25
App/request referral

FrozenKiwi:
Embedded prefinery form for registering sign-up links.

Not sure if we'll actually push this, but we need something

image


apps/site-landing/src/containers/ApplyBeta/index.tsx
```diff
@@ -0,0 +1,69 @@
...
+          <FormattedMessage {...translations.whatToExpectContent} />
+        </Message.Content>
+      </Message>
+      <div className='prefinery-form-embed'>
+      </div>
+      <Helmet>
+        <script src="https://widget.prefinery.com/widget/v2/5jn8p69h.js" defer></script>
+      </Helmet>
+    </div>
+  );
```
FrozenKiwi, line 68: This could clearly be a lot better

FrozenKiwi: The text really needs to be prettied up, but I just couldn't put my head in writer mode when I wrote it.

libs/site-base/lang/en.json
```diff
@@ -235,6 +235,10 @@
     "defaultMessage": "Create Account",
     "description": "app.account.create.form.buttonCreate: Button for the create account form page"
   },
+  "bIBRtw": {
+    "defaultMessage": "Join our Beta",
+    "description": "Super-title on request access page"
```
MarieTheCoin, line 240: Is this a new type of title?

MarieTheCoin (approved):

libs/site-base/lang/en.json
FrozenKiwi: Err... no - just short-hand for "Above the Title".  Looks a little silly now I read it

PR #467, opened 2023-05-22
add ability to cancel incomplete deposits

FrozenKiwi:
Ideally, a cancel button for each/every action type


FrozenKiwi:
NOTE: This should also be used to clean up the cancelled bill payment

PR #905, opened 2026-06-16
Harvester/always log elements

FrozenKiwi:
## Summary by CodeRabbit

* **Bug Fixes**
  * Improved snapshot/MHTML capture reliability with safer cleanup even when capture fails.
  * Ensured MHTML is no longer skipped for section-based progress, and enhanced error capture consistency.
  * Refined estimated text scoring heuristics for more accurate fuzzy matching.

* **Chores**
  * Removed section-skipping configuration from agent/serializer setup and related tooling scripts.
  * Reduced redundant non-error API event logging while still recording all errors.
  * Standardized callback initialization to always enable element screenshot output.


FrozenKiwi:
@coderabbitai review

